PR #203, opened 2021-05-24
RC6

alcueca:
- fix: give permission to the fyToken to transfer base from the join 
 - fix: fyToken burn behaviour
 - Weth is now an immutable variable of Ladle 
 - Symmetrical source set 
 - Get chainlink inverse price 
 - Use AccessControl for MultiOracles 
 - Implement dust limits
 - Flash loan size to debt size as a parameter in `roll`
 - fix: borrowing fees when rolling are applied with the new maturity
 - remove and repay, remove and redeem
 - BUSL license


brucedonovan (commented):
happy with these changes

PR #213, opened 2021-06-17
[WIP] Fix code423n4 #71

uivlis:
[This](https://github.com/code-423n4/2021-05-yield-findings/issues/71) issue may require a solution that would need a new version of @yield-protocol/vault-interfaces. I would have to add a new function to the ICauldron and IFYToken interfaces, respectively, the debtInBaseAndFYToken and accrual functions. I will circle back to this once @albertocuestacanada gives the green light that the current solution is, in fact, right.


uivlis:
Part of the `Witch.sol` refactoring. In progress...

PR #216, opened 2021-06-18
Fix code423n4/70

uivlis:
@albertocuestacanada When you are available, I'd very much like you to help me with the tests for this branch, I've run out of ideas why some of tx's revert.

For [this](https://github.com/code-423n4/2021-05-yield-findings/issues/70).

### DIFF
--- a/contracts/JoinFactory.sol
+++ b/contracts/JoinFactory.sol
@@ -9,9 +9,6 @@ import "./Join.sol";
 contract JoinFactory is IJoinFactory {
 
   /// @dev Deploys a new join.
-  /// The asset address is written to a temporary storage slot to allow for simpler
-  /// address calculation, while still allowing the Join contract to store the values as
-  /// immutable.
   /// @param asset Address of the asset token.
   /// @return join The join address.
   function createJoin(address asset) external override returns (address) {
--- a/contracts/JoinFactory.sol
+++ b/contracts/JoinFactory.sol
@@ -7,53 +7,8 @@ import "./Join.sol";
 
 /// @dev The JoinFactory can deterministically create new join instances.
 contract JoinFactory is IJoinFactory {
-  /// Pre-hashing the bytecode allows calculateJoinAddress to be cheaper, and
-  /// makes client-side address calculation easier
-  bytes32 public constant override JOIN_BYTECODE_HASH = keccak256(type(Join).creationCode);
 
   address private _nextAsset;
-
-  /// @dev Returns true if `account` is a contract.
-  function isContract(address account) internal view returns (bool) {
-      // This method relies on extcodesize, which returns 0 for contracts in
-      // construction, since the code is only stored at the end of the
-      // constructor execution.
-
-      uint256 size;
-      // solhint-disable-next-line no-inline-assembly
-      assembly { size := extcodesize(account) }
-      return size > 0;
-  }
-
-  /// @dev Calculate the deterministic addreess of a join, based on the asset token.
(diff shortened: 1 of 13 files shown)


albertocuestacanada:
Sure, give me a couple of days. This is quite a tricky PR.

albertocuestacanada:
The `Wand` deploys each `Join` during `addAsset`, so in fixtures two Joins were deployed per asset, the first would be registered in the Ladle, and the second in the Joins map.

As a result, we were loading tokens into a join which was not the one that the fyToken was pulling from. The `revert` without message was a panic from [this line](https://github.com/yieldprotocol/vault-v2/blob/4906e9cfd7055eb915d2af8196bed8247ae92f61/contracts/Join.sol#L78).

contracts/JoinFactory.sol
```diff
@@ -63,7 +18,7 @@ contract JoinFactory is IJoinFactory {
   /// @return join The join address.
   function createJoin(address asset) external override returns (address) {
     _nextAsset = asset;
-    Join join = new Join{salt: keccak256(abi.encodePacked(asset))}();
+    Join join = new Join();
```
albertocuestacanada, line 21: Please change the `Join` constructor to accept the `asset` as a parameter, we don't need to use a callback anymore because we don't use CREATE2 anymore.

albertocuestacanada (changes requested):
Simplify JoinFactory by passing `asset` as a parameter to the Join constructor, please.

uivlis:
Ah, you're right, thank you so much!

contracts/JoinFactory.sol
```diff
@@ -7,53 +7,6 @@ import "./Join.sol";
...
-  function getJoin(address asset) external view override returns (address join) {
-    join = _calculateJoinAddress(asset);
-
-    if(!isContract(join)) {
-      join = address(0);
-    }
-  }
 
   /// @dev Deploys a new join.
   /// The asset address is written to a temporary storage slot to allow for simpler
```
albertocuestacanada, line 12: Fix natspec, please

albertocuestacanada (changes requested):
Minor changes now

test/021_join.ts
```diff
@@ -39,9 +41,13 @@ describe('Join', function () {
     token = (await deployContract(ownerAcc, ERC20MockArtifact, ['MTK', 'Mock Token'])) as ERC20Mock
     otherToken = (await deployContract(ownerAcc, ERC20MockArtifact, ['OTH', 'Other Token'])) as ERC20Mock
     joinFactory = (await deployContract(ownerAcc, JoinFactoryArtifact, [])) as JoinFactory
-    const joinAddress = await joinFactory.calculateJoinAddress(token.address) // Get the address
-    await joinFactory.createJoin(token.address) // Create the Join (doesn't return anything outside a contract call)
-    join = (await ethers.getContractAt('Join', joinAddress, ownerAcc)) as Join
+    join = (await ethers.getContractAt(
```
albertocuestacanada, line 44: Could you please extract this code into a helper function in helpers.ts or something like that? It's quite some code being repeated, and quite complex.

uivlis:
Sure!

albertocuestacanada (approved):
I'd rename `sendStatic` as `sendWithReturn` or something a bit more descriptive. Up to you, though. Good job.

uivlis:
I'd kindly ask your permission to keep the `sendStatic` name. I find the contrast between `callStatic` and `sendStatic` not only descriptive but also memorable.

PR #217, opened 2021-06-22
Code423n4/58

uivlis:
Fix for [this](https://github.com/code-423n4/2021-05-yield-findings/issues/58).


albertocuestacanada (changes requested):
It should return `redeemed`, not `amount`.

As a rule of thumb, always return any values that were calculated during a function.

uivlis:
Huh?

albertocuestacanada (approved):
My bad

PR #219, opened 2021-06-23
Code423n4/32

uivlis:
Fix for [this](https://github.com/code-423n4/2021-05-yield-findings/issues/32).

### DIFF
--- a/contracts/FYToken.sol
+++ b/contracts/FYToken.sol
@@ -4,6 +4,7 @@ pragma solidity ^0.8.0;
 import "erc3156/contracts/interfaces/IERC3156FlashBorrower.sol";
 import "erc3156/contracts/interfaces/IERC3156FlashLender.sol";
 import "@yield-protocol/utils-v2/contracts/token/ERC20Permit.sol";
+import "@yield-protocol/utils-v2/contracts/token/SafeERC20Namer.sol";
 import "@yield-protocol/vault-interfaces/IFYToken.sol";
 import "@yield-protocol/vault-interfaces/IJoin.sol";
 import "@yield-protocol/vault-interfaces/IOracle.sol";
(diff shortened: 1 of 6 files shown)


uivlis:
CI fails due to npm down.

albertocuestacanada (changes requested):
SafeERC20Namer must be used as an external library (to reduce deployment costs). You'll need to do changes in tests, fixtures, and environments-v2. Check how it was done for Pool.sol if you need an example.

albertocuestacanada:
What I don't understand is that in yieldspace-v2 I had to use ethers to add SafeERC20Namer as an external library (https://github.com/yieldprotocol/yieldspace-v2/blob/69379de0fbadc9c4b4d56c1057c0fee8f0fc40e3/test/shared/fixtures.ts#L78) to PoolFactory.

Here, how does the FYToken.sol constructor know the deployment address of SafeERC20Namer?

uivlis:
Alright.

I've spent the past 4 hours trying to link this. It's no use (literally)! Like the compiler tells me:
```
NomicLabsHardhatPluginError: You tried to link the contract Wand with SafeERC20Namer, which is not one of its libraries.
This contract doesn't need linking any libraries.
```
Same if I try to link it with `FYToken`. By link I mean:
```
const Wand = await ethers.getContractFactory('FYToken', {
      libraries: {
        SafeERC20Namer: safeERC20Namer.address
      }
    })
```
since by using Waffle's `link` (https://ethereum-waffle.readthedocs.io/en/latest/basic-testing.html#linking) (which would normally fit into the `deployContract` style of the tests), I went down the rabbit hole pretty fast. All sorts of new and irrelevant things I've learned, like how Buidelr's compiled JSON's have a different standard than those of Waffle's, and how linking for solidityV5 was changed from solidityV4, etc., just to end up being recommended to use hardhat's factory linking.

I mean I'm not sure what's the issue here, but I guess if the library wouldn't have been previously linked, the tests wouldn't have passed, right?

If you'd like to have a try, I'd be more than interested in how you would spin this, but, as for me, I can't do anything more, unfortunately.

PR #220, opened 2021-06-28
Clean builds and rebuild before tests

uivlis:
I've found this a very handy fix. Basically, removes the `artifacts` and `typechain` folders before compiling for tests. Not linked to any particular issue.


albertocuestacanada (approved):
Looks good to me!

Issue #214, opened 2021-06-17
Refactor Witch


albertocuestacanada:
Issues:
1. Parallel liquidation engines imply multiple `grab` which are complex.
2. No DUST parameter
3. Debt accrual not taken into account
4. Need for a `payAll` function.

albertocuestacanada:
1. Remove multiple liquidation engines, and with it the auction interval. As a result, move the auction timestamps to Witch. Vaults are now only released when their debt becomes zero.
2. Introduce a DUST parameter for liquidations. A `buy` that leaves collateral under DUST in the Witch reverts.
3. In `buy`:
 - Rename the `art` parameter as `base`.
 - Calculate `art` from `base`, using `cauldron.debtFromBase`
 - Remains unchanged:
```
        {
            // Price of a collateral unit, in underlying, at the present moment, for a given vault
            //
            //                ink                     min(auction, elapsed)
            // price = 1 / (------- * (p + (1 - p) * -----------------------))
            //                art                          auction
            (uint256 auctionTime_, uint256 initialProportion_) = (auctionTime, initialProportion);
            uint256 term1 = uint256(balances_.ink).wdiv(balances_.art);
            uint256 dividend2 = auctionTime_ < elapsed ? auctionTime_ : elapsed;
            uint256 divisor2 = auctionTime_;
            uint256 term2 = initialProportion_ + (1e18 - initialProportion_).wmul(dividend2.wdiv(divisor2));
            price = uint256(1e18).wdiv(term1.wmul(term2));
        }
        uint256 ink = uint256(art).wdivup(price);  
```
 - Settle with `base`: `ladle.settle(vaultId, msg.sender, ink.u128(), base);`
 - Call `slurp` from Witch. Converting again between base and fyToken in `settle` would be inefficient. On the other hand, now `settle` has more room to be abused.
4. Create `Witch.buyAll` that buys all the collateral from a vault in liquidation for an undetermined amount of base. Does it need two slippage guards (minInk, maxBase)?

Consider breaking this up into 4 issues.

albertocuestacanada:
[Done](https://github.com/yieldprotocol/vault-v2/pull/265)

PR #222, opened 2021-07-08
Rc7

albertocuestacanada:
Fixes from the Code Arena review.
 - Check seriesId, ilkId, and vaultId can't be zero
 - Random vault ids
 - Block `stir` to the same address
 - Replace `public auth` with `external auth`
 - Make JoinFactory use `CREATE` and `wand.addAsset` deploys the Join
 - Only receive Ether from WETH
 - PoolRouter.batch returns the return values of the batch actions
 - Check for function signature collisions in CI

### DIFF
--- a/contracts/Cauldron.sol
+++ b/contracts/Cauldron.sol
@@ -8,6 +8,7 @@ import "./math/WMul.sol";
 import "./math/WDiv.sol";
 import "./math/CastU128I128.sol";
 import "./math/CastI128U128.sol";
+import "./math/CastU256U128.sol";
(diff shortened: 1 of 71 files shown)


albertocuestacanada:
Not sure why the last merge breaks the tests, and reverting it doesn't seem to work. I'm branching off instead from the last working commit.

albertocuestacanada:
That didn't fix it. It seems the error happens when you run 031 and 051 in sequence, so it might have to do with some corruption of the state when moving the time. Will fix later.

albertocuestacanada:
Fixes #214

PR #224, opened 2021-07-08
Witch refactor

### LINKED ISSUE #214: Refactor Witch

albertocuestacanada:
Fixes #214 with a refactor of Witch, and with it a number of issues in the C4 report ([8](https://github.com/code-423n4/2021-05-yield-findings/issues/8), [44](https://github.com/code-423n4/2021-05-yield-findings/issues/44), [71](https://github.com/code-423n4/2021-05-yield-findings/issues/71))

 - Removed parallel liquidation engines feature, which was complex and a source of exploits.
 - Included a DUST parameter
 - Debt accrual now taken into account for liquidations
 - payAll function to liquidate a whole vault after maturity

There is an issue unrelated to this PR making CI fail. I believe it has something to do with the chain not being reverted to the previous block after advancing time in the 031 test file.

### DIFF
--- a/contracts/Witch.sol
+++ b/contracts/Witch.sol
@@ -89,7 +89,7 @@ contract Witch is AccessControl() {
         {
             uint256 elapsed = uint32(block.timestamp) - auction_.start;                      // Auctions will malfunction on the 7th of February 2106, at 06:28:16 GMT, we should replace this contract before then.
             uint256 price = inkPrice(balances_, initialOffer_, duration_, elapsed);
-            ink = uint256(art).wdivup(price);                                                    // Calculate collateral to sell. Using divdrup stops rounding from leaving 1 stray wei in vaults.
+            ink = uint256(art).wmul(price);                                                    // Calculate collateral to sell. Using divdrup stops rounding from leaving 1 stray wei in vaults.
             require (ink >= min, "Not enough bought");
             require (ink == balances_.ink || balances_.ink - ink >= dust_, "Leaves dust");
         }
@@ -119,7 +119,7 @@ contract Witch is AccessControl() {
         {
             uint256 elapsed = uint32(block.timestamp) - auction_.start;                      // Auctions will malfunction on the 7th of February 2106, at 06:28:16 GMT, we should replace this contract before then.
             uint256 price = inkPrice(balances_, initialOffer_, duration_, elapsed);
-            ink = uint256(balances_.art).wdivup(price);                                                    // Calculate collateral to sell. Using divdrup stops rounding from leaving 1 stray wei in vaults.
+            ink = uint256(balances_.art).wmul(price);                                                    // Calculate collateral to sell. Using divdrup stops rounding from leaving 1 stray wei in vaults.
             require (ink >= min, "Not enough bought");
             require (ink == balances_.ink || balances_.ink - ink >= dust_, "Leaves dust");
         }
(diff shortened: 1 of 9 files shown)


contracts/Witch.sol
```diff
@@ -9,79 +9,141 @@ import "./math/WMul.sol";
...
-    uint128 public auctionTime = 4 * 60 * 60; // Time that auctions take to go to minimal price and stay there.
-    uint128 public initialProportion = 5e17;  // Proportion of collateral that is sold at auction start.
+    struct Auction {
+        address owner;
+        uint32 start;
+    }
+
+    uint32 public duration = 4 * 60 * 60; // Time that auctions take to go to minimal price and stay there.
+    uint64 public initialOffer = 5e17;  // Proportion of collateral that is sold at auction start (1e18 = 100%)
+    uint128 public dust;                     // Minimum collateral that must be left when buying, unless buying all
```
aniemerg, line 35: This the same for all collateral types? That doesn't seem right.

contracts/Witch.sol
```diff
@@ -9,79 +9,141 @@ import "./math/WMul.sol";
...
+            ink = uint256(art).wdivup(price);                                                    // Calculate collateral to sell. Using divdrup stops rounding from leaving 1 stray wei in vaults.
+            require (ink >= min, "Not enough bought");
+            require (ink == balances_.ink || balances_.ink - ink >= dust_, "Leaves dust");
         }
-        uint256 ink = uint256(art).wdivup(price);                                                    // Calculate collateral to sell. Using divdrup stops rounding from leaving 1 stray wei in vaults.
-        require (ink >= min, "Not enough bought");
 
-        ladle.settle(vaultId, msg.sender, ink.u128(), art);                                        // Move the assets
+        cauldron.slurp(vaultId, ink.u128(), art.u128());                                            // Remove debt and collateral from the vault
+        ladle.settle(vaultId, msg.sender, ink.u128(), base);                                        // Move the assets
```
aniemerg, line 98: It is somewhat concerning to me that the Ladle is a dependency of Witch. What happens if we need to upgrade Ladle, can it cause issues with already operating auctions? If we upgrade Ladle, what is the likelihood that we need to update Witch?

aniemerg (approved):

contracts/Witch.sol
albertocuestacanada: Would only dust have to be different for all collateral types, or also the auction duration and the initial proportion of the collateral that is offered?

contracts/Witch.sol
albertocuestacanada: Right now, if the Ladle is replaced, the Witch must be replaced as well. The process to let the old Witch finish ongoing auctions would be as follows:

1. New Ladle and new Witch are deployed, but not permissioned yet (so that everything fits in a transaction).
2. New Ladle and new Witch are configured.
3. New Ladle and new Witch are permissioned, allowing them to start operations immediately. Old Witch is removed  from `Cauldron.grab` permissions to stop new auctions being initiated with it. The old Ladle is removed from all Cauldron permissions, which still allows the old Witch to finish ongoing auctions using `ladle.settle`.
4. Once all auctions in the old Witch are finished, the `ladle.settle` permissions are removed. The old Ladle is removed from the permissions of all the Joins.

I will remove the Ladle dependency from the Witch, though. The dependency is there because I was concerned of governance errors in setting up new assets (which would need to be set up on both the Ladle and the Witch). Using the Wand, though, the possibility of error is removed.

Issue #237, opened 2021-07-12
Remove `ladle.settle`


albertocuestacanada:
Instead, assets should be set up through the `Wand` so that the possibility of error in adding them to the Ladle but not the Witch is removed.

albertocuestacanada:
[Done](https://github.com/yieldprotocol/vault-v2/pull/265)

PR #242, opened 2021-07-13
Tweak/remove settle

albertocuestacanada:
By setting up the Witch to liquidate assets when these are set up in the Wand, we can safely remove its dependency from Ladle.


albertocuestacanada:
Fixes #237

aniemerg (approved):

Issue #238, opened 2021-07-12
Set up `Witch.dust` per collateral type


albertocuestacanada:
Also auction duration and initialOffer?

albertocuestacanada:
Done in RC8

PR #243, opened 2021-07-13
feat: configure witch parameters per ilk

albertocuestacanada:
Now the Witch stores different duration, initialOffer and dust for each collateral that can be liquidated.


albertocuestacanada:
Fixes #238

aniemerg (approved):

PR #245, opened 2021-07-14
Composite oracles

albertocuestacanada:
Composite oracles store other IOracle contracts as sources, and for any given pair of (base, quote) can also store a path.

When (base, quote) is requested, the Composite Oracle checks if it has a path, and if so it returns the product of all (base, quote) pairs as in (base, path0) * (path0, path1) * ... * (pathN-1 pathN) * (pathN, quote).

### DIFF
--- a/contracts/oracles/composite/CompositeMultiOracle.sol
+++ b/contracts/oracles/composite/CompositeMultiOracle.sol
@@ -1,0 +1,147 @@
+// SPDX-License-Identifier: BUSL-1.1
+pragma solidity 0.8.1;
+
+import "@yield-protocol/utils-v2/contracts/access/AccessControl.sol";
+import "@yield-protocol/vault-interfaces/IOracle.sol";
+import "../../math/CastBytes32Bytes6.sol";
+
+
+/**
+ * @title CompositeMultiOracle
+ */
+contract CompositeMultiOracle is IOracle, AccessControl {
+    using CastBytes32Bytes6 for bytes32;
+
+    uint8 public constant override decimals = 18;   // All prices are converted to 18 decimals
+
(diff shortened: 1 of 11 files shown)


contracts/oracles/composite/CompositeMultiOracle.sol
```diff
@@ -0,0 +1,147 @@
...
+                uint256 updateTime_;
+                (price_, updateTime_) = IOracle(source.source).peek(base_, quote, 10 ** source.decimals);   // Get price for one unit
+                price = price * price_ / (10 ** source.decimals);                                           // Fixed point according to decimals
+                updateTime = (updateTime > updateTime_) ? updateTime : updateTime_;                         // Take the oldest update time
+                base_ = path[p];
+            }
+        }
+    }
+
+    function _get(bytes6 base, bytes6 quote) private returns (uint price, uint updateTime) {
```
aniemerg, line 113: Where is this used?

aniemerg (approved):

contracts/oracles/composite/CompositeMultiOracle.sol
albertocuestacanada: Some oracles have both transactional and non-transactional price feeds. For example, taking a cToken as a source for a price feed, you can use `exchangeRateCurrent` or `exchangeRateStored`.

The Cauldron always retrieves prices using transactional feeds (`get`), although then the rate and chi oracle uses the non-transactional function (`_peek`) under the hood to save gas. The Chainlink spot oracles also converge to the non-transactional function (`_peek`).

The only case where the Cauldron ends using a transaction price feed is when using a cToken as a collateral, because we want to end using `exchangeRateCurrent` in that case.

albertocuestacanada:
The checks will pass once I merge stuff and update interfaces

PR #246, opened 2021-07-15
feat: cToken oracles

albertocuestacanada:
To accept cToken as a collateral for "Token" all that we need is an IOracle that returns the `exchangeRate` from the cToken as the cToken/Token spot price.

This can be used in a composite oracle to accept cTokenX as a collateral for tokenY.

### DIFF
--- a/contracts/FYTokenFactory.sol
+++ b/contracts/FYTokenFactory.sol
@@ -1,0 +1,44 @@
+// SPDX-License-Identifier: BUSL-1.1
+pragma solidity 0.8.1;
+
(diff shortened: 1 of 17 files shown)


albertocuestacanada:
As before, the tests will pass once I merge RC8 into this branch. I could do it now, but it would be harder to review.

albertocuestacanada:
All checks will pass once I merge rc8 into this PR. I won't do it yet because it would make it harder to review.

aniemerg (approved):

PR #251, opened 2021-08-09
draft: TWAR oracle

alcueca:
A simple IOracle contract that returns the time-weighted average ratio in a YieldSpace pool, measured in ratioSeconds.

Not really the usual IOracle, since it doesn't make sense to multiply a ratio by an amount, it wouldn't return the fyToken that you can buy with a given base amount. It is useful to track pool ratios and avoid being sandwiched.


devtooligan:
closing -- tracking on ticket https://github.com/yieldprotocol/vault-v2/issues/336

PR #258, opened 2021-08-12
Feat/router

albertocuestacanada:
This PR reorganizes how the Ladle deals with routing arbitrary `call`, as well as `transfer` and `permit`. The goal is to make the Ladle safer, while allowing a few more recipes to be implemented.

Originally, arbitrary calls could be routed to Pools. This is problematic in principle because the Ladle would be doing arbitrary calls to a contract, and the Ladle has permission to do dangerous things via `auth` and ERC20 `approve`. The issue was mitigated until now by whitelisting the `route` targets via `addPool`.

Thinking on how to `route` calls to other contracts, such as Strategy.sol, revealed the risk of letting the Ladle do arbitrary calls, chosen by the user. Even if the targets are whitelisted, it is easy to miss interactions that would lead to loss of funds.

The solution in this PR is to use a Router contract. This Router contract is deployed automatically when the Ladle is deployed, and all it does is to take arbitrary calls to arbitrary targets from Ladle and execute them. This solves the problem of the Ladle having permissions, since it will always be the Router the actual caller, and the Router will never have any permissions by our hand.

In a way, it is an airgap that for the cost of one extra CALL (no SLOADS even), strips all permissions from Ladle before it calls on a contract. In any case, targets for this kind of execution are still whitelisted as an additional layer of security.

The whitelisting has been extended to `transfer` and `permit`, to allow the Ladle to transfer any whitelisted tokens from the users wallet to any receiver. The main protection here is that `transfer` only takes tokens from the caller, so it can't ever be used to steal funds from an user's wallet. Still, token that Ladle will cal `transfer` or `permit` on are whitelisted as an additional layer of security.

The `addJoin` and `addPool` governance functions have been extended to whitelist standard assets and pools appropriately and automatically, meaning that the governance UX remains the same. Only now governance has more options to whitelist tokens and integrations.

The syntax for certain batch actions has changed for the better. Where before the frontend needed to use asset ids, boolean flags, or combinations of addresses to infer the actual target, now a single address can be passed on.


aniemerg:
Is this approach adding a lot of gas to a set of calls?

albertocuestacanada:
A `call` routed through the `Router` (and therefore stripped of permissions) adds about 2300K gas, or 1 SLOAD. At most there are two such `route` calls in any of our most complex batches.

PR #290, opened 2021-09-19
add: Aave Lending Oracle

Aveesh52:


alcueca:
Maybe the issue wasn't very clear, but the AaveLendingOracle is supposed to be a Yield v2 contract that returns the borrowing and saving rates from Aave in the same format as our [CompoundMultiOracle](https://github.com/yieldprotocol/vault-v2/blob/rc11/contracts/oracles/compound/CompoundMultiOracle.sol).

In Compound, the borrowing rate is returned by `CTokenInterface(source).borrowIndex()`, and the savings rate is the exchange rate between a cToken and its underlying. I don't know how that works in Aave.

```
        if (kind == RATE.b6()) accumulator = CTokenInterface(source).borrowIndex();
        else if (kind == CHI.b6()) accumulator = CTokenInterface(source).exchangeRateStored();
```

It seems to me that you took the [contract that manages the Chainlink oracles in Aave](https://github.com/aave/aave-protocol/blob/master/contracts/misc/ChainlinkProxyPriceProvider.sol) and copied it over.

Aveesh52:
there is no exchange rate in Aave
1 aDAI => 1 DAI , For borrowing rate , there's getReserveData we ca use
https://docs.aave.com/developers/guides/apy-and-apr

alcueca:
It's more likely that we'll implement our own rate accumulators than this.

PR #308, opened 2021-10-31
feat: lido oracle

iamsahu:
**Purpose**
Testing & building Lido Oracle for supporting WstETH as a collateral

**Details**
1. A lido oracle was implemented following the IOracle pattern to fetch WstETH/stETH exchange rate.
2. A composite oracle for the path was WstETH->stETH->ETH was added and tested

**Code to run**
013_oracles_composite.ts
016_oracle_lido.ts

**Test Output**

```
Praffuls-MacBook-Pro:vault-v2 praffulsahu$ npx hardhat test test/013_oracles_composite.ts 
  Oracles - Composite
    ✓ retrieves the value at spot price for direct pairs (152ms)
    ✓ retrieves the value at spot price for DAI -> USDC and reverse (68ms)
    Lido Oracle
      ✓ retrieves the value at spot price for direct pairs (70ms)
      ✓ retrieves the value at spot price for WSTETH -> ETH and reverse (52ms)
  4 passing (9s)
```

```
Praffuls-MacBook-Pro:vault-v2 praffulsahu$ npx hardhat test test/016_oracle_lido.ts
  Oracles - Chainlink
    ✓ sets and retrieves the value at spot price
    ✓ revert on unknown sources
2 passing (1s)
```


contracts/mocks/oracles/lido/LidoMock.sol
```diff
@@ -0,0 +1,19 @@
...
+
+contract LidoMock is ISourceMock {
+    uint256 public price;
+
+    function set(uint256 price_) external override {
+        price = price_;
+    }
+
+    function getWstETHByStETH(uint256 _stETHAmount) external view returns (uint256) {
+        return (_stETHAmount * 10**18) / price;
```
alcueca, line 13: It's just a mock, but the `1e18` notation is cheaper in gas costs than `10**18`.

alcueca (changes requested):
Great job, all my comments but two were quite minor ones about natspec or variable naming. The important ones:
1. Pass the assetIds on the constructor, to be able to choose the wstETH and stETH ids on deployment.
2. Revert if passed on asset ids different from wstEthId or stEthId

alcueca:
@richiehumphrey, please have a look at this PR as well, you'll need to do a spot oracle as well.

richiehumphrey (approved):
Looks great!  Super tiny natspec typos is all I found

There are some failing tests we should probably fix before merge, looks like it might be missing a steth mock?

contracts/oracles/lido/LidoOracle.sol
```diff
@@ -0,0 +1,68 @@
...
+import '../../constants/Constants.sol';
+import './IWstETH.sol';
+
+/**
+ * @title LidoOracle
+ * @notice
+ */
+contract LidoOracle is IOracle, AccessControl, Constants {
+    using CastBytes32Bytes6 for bytes32;
+    IWstETH wstETH;
```
richiehumphrey, line 17: should we explicitly set the visibility here?

contracts/oracles/lido/LidoOracle.sol
```diff
@@ -0,0 +1,68 @@
...
+        uint256 baseAmount
+    ) external virtual override returns (uint256 quoteAmount, uint256 updateTime) {
+        return _peek(base.b6(), quote.b6(), baseAmount);
+    }
+
+    function _peek(
+        bytes6 base,
+        bytes6 quote,
+        uint256 baseAmount
+    ) private view returns (uint256 quoteAmount, uint256 updateTime) {
```
richiehumphrey, line 56: I doubt this will ever have any child contracts, but should we follow the practice of using internal by default?

`V10 - Only use private to intentionally prevent child contracts from accessing the variable, prefer internal for flexibility.`

contracts/oracles/lido/LidoOracle.sol
```diff
@@ -0,0 +1,68 @@
...
+    }
+
+    /**
+     * @notice Retrieve the value of the amount at the latest oracle price. Same as `peek` for this oracle.
+     */
+    function get(
+        bytes32 base,
+        bytes32 quote,
+        uint256 baseAmount
+    ) external virtual override returns (uint256 quoteAmount, uint256 updateTime) {
```
richiehumphrey, line 48: Is there a reason we are using virtual here?

contracts/oracles/lido/LidoOracle.sol
```diff
@@ -0,0 +1,68 @@
...
+    IWstETH wstETH;
+    bytes32 public immutable wstEthId;
+    bytes32 public immutable stEthId;
+
+    constructor(bytes32 wstEthId_, bytes32 stEthId_) public {
+        wstEthId = wstEthId_;
+        stEthId = stEthId_;
+    }
+
+    function setSource(IWstETH wstETH_) external auth {
```
richiehumphrey, line 26: natspec?

contracts/oracles/lido/LidoOracle.sol
```diff
@@ -0,0 +1,68 @@
...
+import '@yield-protocol/utils-v2/contracts/token/IERC20Metadata.sol';
+import '@yield-protocol/vault-interfaces/IOracle.sol';
+import '../../constants/Constants.sol';
+import './IWstETH.sol';
+
+/**
+ * @title LidoOracle
+ * @notice
+ */
+contract LidoOracle is IOracle, AccessControl, Constants {
```
richiehumphrey, line 15: in general the natspec on this contract is a little light, although I realize there probably isn't much explanation necessary and that its based on other Yield oracles

`T6 - Is the purpose of the contract and how it interacts with others documented using natspec?`
`T12 - Summarize the purpose and functionality of the contract with a @notice natspec comment. Document how the contract interacts with other contracts inside/outside the project in a @dev natspec comment.`
`V5 - Is the purpose of the variable and other important information documented using natspec?`
`F15 - Are all arguments, return values, side effects and other information documented using natspec?`

contracts/oracles/lido/LidoOracle.sol
```diff
@@ -0,0 +1,68 @@
...
+    bytes32 public immutable wstEthId;
+    bytes32 public immutable stEthId;
+
+    constructor(bytes32 wstEthId_, bytes32 stEthId_) public {
+        wstEthId = wstEthId_;
+        stEthId = stEthId_;
+    }
+
+    function setSource(IWstETH wstETH_) external auth {
+        wstETH = wstETH_;
```
richiehumphrey, line 27: Should we emit an event here?
`T2 - Are events emitted for every storage mutating function?`

contracts/oracles/lido/LidoOracle.sol
```diff
@@ -0,0 +1,68 @@
...
+    function _peek(
+        bytes6 base,
+        bytes6 quote,
+        uint256 baseAmount
+    ) private view returns (uint256 quoteAmount, uint256 updateTime) {
+        require((base == wstEthId && quote == stEthId) || (base == stEthId && quote == wstEthId), 'Source not found');
+
+        if (base == wstEthId) {
+            //Base equals WstETH
+            quoteAmount = wstETH.getStETHByWstETH(baseAmount);
```
richiehumphrey, line 61: Is there any need to handle the possibility of getting back some weird or erroneus data?

`X5 - Is the result checked and errors dealt with? (SWC-104)`

contracts/oracles/lido/LidoOracle.sol
```diff
@@ -0,0 +1,68 @@
...
+ * @title LidoOracle
+ * @notice
+ */
+contract LidoOracle is IOracle, AccessControl, Constants {
+    using CastBytes32Bytes6 for bytes32;
+    IWstETH wstETH;
+    bytes32 public immutable wstEthId;
+    bytes32 public immutable stEthId;
+
+    constructor(bytes32 wstEthId_, bytes32 stEthId_) public {
```
richiehumphrey, line 21: I ran Slither on this contract and the only hit was this:
```
Warning: Visibility for constructor is ignored. If you want the contract to be non-deployable, making it "abstract" is sufficient.
  --> contracts/oracles/lido/LidoOracle.sol:21:5:
   |
21 |     constructor(bytes32 wstEthId_, bytes32 stEthId_) public {
   |     ^ (Relevant source part starts here and spans across multiple lines).


```

contracts/oracles/lido/LidoOracle.sol
```diff
@@ -0,0 +1,68 @@
...
+    }
+
+    /**
+     * @notice Retrieve the value of the amount at the latest oracle price.
+     */
+    function peek(
+        bytes32 base,
+        bytes32 quote,
+        uint256 baseAmount
+    ) external view virtual override returns (uint256 quoteAmount, uint256 updateTime) {
```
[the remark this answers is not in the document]
richiehumphrey: also : 
Is there a reason we are using virtual here?
`F19 - Use virtual if there are legitimate (and safe) instances where a child contract may wish to override the function's behavior.`

test/013_oracles_composite.ts
```diff
@@ -1,24 +1,30 @@
...
+import { ETH, DAI, USDC, WSTETH, STETH } from '../src/constants'
 
 import ChainlinkMultiOracleArtifact from '../artifacts/contracts/oracles/chainlink/ChainlinkMultiOracle.sol/ChainlinkMultiOracle.json'
 import CompositeMultiOracleArtifact from '../artifacts/contracts/oracles/composite/CompositeMultiOracle.sol/CompositeMultiOracle.json'
 import ChainlinkAggregatorV3MockArtifact from '../artifacts/contracts/mocks/oracles/chainlink/ChainlinkAggregatorV3Mock.sol/ChainlinkAggregatorV3Mock.json'
 import DAIMockArtifact from '../artifacts/contracts/mocks/DAIMock.sol/DAIMock.json'
 import USDCMockArtifact from '../artifacts/contracts/mocks/USDCMock.sol/USDCMock.json'
 import WETH9MockArtifact from '../artifacts/contracts/mocks/WETH9Mock.sol/WETH9Mock.json'
-
+import STETHMockArtifact from '../artifacts/contracts/mocks/STETHMock.sol/STETHMock.json'
```
richiehumphrey, line 12: This is so weird because in the other test it doesn't have a problem finding it!

contracts/oracles/lido/LidoOracle.sol
```diff
@@ -0,0 +1,68 @@
...
+    ) external virtual override returns (uint256 quoteAmount, uint256 updateTime) {
+        return _peek(base.b6(), quote.b6(), baseAmount);
+    }
+
+    function _peek(
+        bytes6 base,
+        bytes6 quote,
+        uint256 baseAmount
+    ) private view returns (uint256 quoteAmount, uint256 updateTime) {
+        require((base == wstEthId && quote == stEthId) || (base == stEthId && quote == wstEthId), 'Source not found');
```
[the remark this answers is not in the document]
alcueca: Yes, we can start using custom errors now.

contracts/oracles/lido/LidoOracle.sol
alcueca: How would you recognize erroneous data?

test/013_oracles_composite.ts
alcueca: Sometimes, after a contract being renamed, the local artifacts still have files that make it work, but the CI copy doesn't work because it's fresh. Removing the artifacts folder in the local copy sometimes reproduces the problem.

test/013_oracles_composite.ts
```diff
@@ -101,4 +111,56 @@ describe('Oracles - Composite', function () {
     expect((await compositeMultiOracle.peek(bytes6ToBytes32(DAI), bytes6ToBytes32(USDC), WAD))[0]).to.equal(oneUSDC)
     expect((await compositeMultiOracle.peek(bytes6ToBytes32(USDC), bytes6ToBytes32(DAI), oneUSDC))[0]).to.equal(WAD)
   })
+
```
alcueca, line 114: I would move the testing of the Lido oracle to it's own testing file (`016_oracles_lido.ts`) and leave the composite test file to just do basic tests on the composite token.

contracts/oracles/lido/LidoOracle.sol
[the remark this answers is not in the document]
richiehumphrey: Nice!!!

@iamsahu [This is the comment](https://github.com/yieldprotocol/vault-v2/pull/305#discussion_r726696088) that started this conversation

contracts/oracles/lido/LidoOracle.sol
richiehumphrey: @alcueca 0 I guess would prob be erroneous?

test/013_oracles_composite.ts
```diff
@@ -101,4 +111,56 @@ describe('Oracles - Composite', function () {
...
+  describe('Lido Oracle', () => {
```
alcueca, line 115: These test are redundant, although I understand you wanting to do them to make sure the combination works. I would move them to `016_oracles_lido.ts`, though.

Also, please rename `016_oracle_lido.ts` to `016_oracles_lido.ts` for consistency

test/033_wsteth.ts
```diff
@@ -0,0 +1,248 @@
+import { SignerWithAddress } from '@nomiclabs/hardhat-ethers/dist/src/signer-with-address'
```
alcueca, line 1: Ladle tests are in the `06*` and `07*` group, and here I think we will be testing the wstEth wrapping module, so I would choose the `074_ladle_wsteth.ts` file name.

I think we will need to create subfolders in the `test` folder, in a separate PR.

test/033_wsteth.ts
```diff
@@ -0,0 +1,248 @@
...
+      'FYToken',
+      (await cauldron.series(seriesId)).fyToken,
+      ownerAcc
+    )) as FYToken
+
+    const pool = (await ethers.getContractAt('PoolMock', await env.ladle.pools(seriesId), ownerAcc)) as PoolMock
+    await pool.mint(ownerAcc.address, true, 0)
+    await fyToken.grantRole(id(fyToken.interface, 'mint(address,uint256)'), ownerAcc.address) // Only test environment
+    await fyToken.mint(pool.address, WAD.mul(2100000))
+    await pool.sync()
```
alcueca, line 190: You can skip the pool initialization if you use a series that was deployed by `fixtures`

test/033_wsteth.ts
```diff
@@ -0,0 +1,248 @@
...
+    // Pour the amount into the pool & borrow
+    expect(await env.ladle.pour(vaultId, ownerAcc.address, parseEther('0.5'), parseEther('0.000001'))).to.emit(
+      cauldron,
+      'VaultPoured'
+    )
+
+    expect(await fyToken['balanceOf(address)'](ownerAcc.address)).to.eq(parseEther('0.000001'))
+  })
+
+  it('Build & test USDC borrowing', async () => {
```
alcueca, line 210: You only need `wand.makeIlk` and `cauldron.addIlks` to get setup

alcueca (changes requested):
Let's have a call for me to walk you through the protocol deployment and setup process. Today in about 4 hours if you are available.

It's a complicated process, it's best if we do it together the first time.

test/033_wsteth.ts
```diff
@@ -0,0 +1,248 @@
...
+    )
+
+    // Setting up Lido Oracle
+    lidoMock = (await deployContract(ownerAcc, LidoMockArtifact)) as LidoMock
+    lidoOracle = (await deployContract(ownerAcc, LidoOracleArtifact)) as LidoOracle
+    await lidoOracle.grantRole(id(lidoOracle.interface, 'setSource(address)'), owner)
+    await lidoOracle['setSource(address)'](lidoMock.address) //mockOracle
+    await lidoMock.set('1008339308050006006')
+
+    // Setting up stethEth Chainlink oracle
```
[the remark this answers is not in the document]
iamsahu: I found the below price feed for stETH/ETH 

https://etherscan.io/address/0x86392dC19c0b719886221c78AB11eb8Cf5c52812

It is taken from the documentation here: https://docs.lido.fi/deployed-contracts#price-feeds

contracts/oracles/lido/LidoOracle.sol
alcueca: I'm never going to say no to good natspec.

contracts/oracles/lido/LidoOracle.sol
alcueca: Ah yeah, `public` was only used in constructors before solidity 0.8

contracts/oracles/lido/LidoOracle.sol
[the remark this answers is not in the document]
alcueca: I think we are using `virtual` as a kind of default here. It's difficult to say whether a contract will be parent to another in the future.

contracts/oracles/lido/LidoOracle.sol
alcueca: Let me ask the twitterverse about that one.

contracts/oracles/lido/LidoOracle.sol
alcueca: `internal` would do just fine, although this is a problem common to all of our oracles, I'm noticing now.

I would leave it as `private` for consistency, and then swap it in all oracles for `internal` in a separate PR

contracts/oracles/lido/LidoOracle.sol
alcueca: Not if you pass 0 as a parameter.

I don't think that you can verify the validity of the data received as a general rule. It seems to me a case-by-case basis, where you know that a function could return wrong data instead of reverting. You can see examples of this defensive programming pattern in TransferHelper, for example.

I think that this kind of defense should only be put up when we can't know in advance who will be the contract that we are calling. I.e., when we let users determine the called contract address by letting them use a factory or something.

alcueca (approved):
Let's ship!

PR #311, opened 2021-11-05
feat: lido module

iamsahu:
**Purpose**
A simple ladle module to handle wrapping & unwrapping of stETH.

**Code**
LidoModule.sol


contracts/modules/LidoModule.sol
```diff
@@ -0,0 +1,58 @@
...
+contract LidoModule is LadleStorage {
+    using WMul for uint256;
+    using CastU256U128 for uint256;
+    using CastU256I128 for uint256;
+    using CastU128I128 for uint128;
+    using TransferHelper for IERC20;
+    using TransferHelper for IWstETH;
+    using TransferHelper for IWETH9;
+    using TransferHelper for address payable;
+    IWstETH public wstETH;
```
alcueca, line 25:
```suggestion
    IWstETH public immutable wstETH;
    IERC20 public immutable stETH;
```

State variables in modules need to be `immutable` to be accessible from within the module. This is because they then get stored in the bytecode, same as functions. Non-immutable state variables get stored in storage, which is not accessible from within a module (because we are using the storage of the contract using `delegatecall`, the Ladle).

contracts/modules/LidoModule.sol
```diff
@@ -0,0 +1,58 @@
...
+
+    constructor(
+        ICauldron cauldron,
+        IWETH9 steth,
```
alcueca, line 29:
```suggestion
        IWETH9 weth,
```
I can't think of exactly how this would break. Maybe it wouldn't. However, it leads to unclear code.

contracts/modules/LidoModule.sol
```diff
@@ -0,0 +1,58 @@
...
+        IWstETH wstETH_
```
alcueca, line 30:
```suggestion
        IWstETH wstETH_
        IERC20 stETH_
```

contracts/modules/LidoModule.sol
```diff
@@ -0,0 +1,58 @@
...
+    ) LadleStorage(cauldron, steth) {
+        wstETH = wstETH_;
```
alcueca, line 32:
```suggestion
        wstETH = wstETH_;
        stETH = stETH_;
```

contracts/modules/LidoModule.sol
```diff
@@ -0,0 +1,58 @@
...
+    function getJoin(bytes6 assetId) internal view returns (IJoin join) {
+        join = joins[assetId];
+        require(join != IJoin(address(0)), 'Join not found');
+    }
+
+    /// @dev Accept stEth, wrap it and forward it to the WstethJoin
+    /// This function should be called first in a batch, and the Join should keep track of stored reserves
+    /// Passing the id for a join that doesn't link to a contract implemnting IWETH9 will fail
+    function joinStEth(bytes6 wstEthId) external returns (uint256 stEthTransferred) {
+        stEthTransferred = weth.balanceOf(address(this));
```
alcueca, line 45:
```suggestion
        stEthTransferred = stEth.balanceOf(address(this));
```

contracts/modules/LidoModule.sol
```diff
@@ -0,0 +1,58 @@
...
+        IJoin wstEthJoin = getJoin(wstEthId);
+        uint256 wrappedAmount = wstETH.wrap(stEthTransferred);
+        wstETH.safeTransfer(address(wstEthJoin), wrappedAmount);
+    }
+
+    /// @dev Unwrap WstETH held by this Ladle, and send the stETH
+    /// This function should be called last in a batch, and the Ladle should have no reason to keep an WSTETH balance
+    function exitStEth(address to) external returns (uint256 ethTransferred) {
+        ethTransferred = wstETH.balanceOf(address(this));
```
alcueca, line 54:
```suggestion
        wstEthTransferred = wstETH.balanceOf(address(this));
```

contracts/modules/LidoModule.sol
```diff
@@ -0,0 +1,58 @@
...
+        uint256 unwrappedAmount = wstETH.unwrap(ethTransferred);
```
alcueca, line 55:
```suggestion
        uint256 unwrappedAmount = wstETH.unwrap(wstEthTransferred);
```

contracts/modules/LidoModule.sol
```diff
@@ -0,0 +1,58 @@
...
+        weth.safeTransfer(to, unwrappedAmount);
```
alcueca, line 56:
```suggestion
        stEth.safeTransfer(to, unwrappedAmount);
```

contracts/modules/LidoModule.sol
```diff
@@ -0,0 +1,58 @@
...
+        wstETH = wstETH_;
+    }
+
+    /// @dev Obtains a join by assetId, and verifies that it exists
+    function getJoin(bytes6 assetId) internal view returns (IJoin join) {
+        join = joins[assetId];
+        require(join != IJoin(address(0)), 'Join not found');
+    }
+
+    /// @dev Accept stEth, wrap it and forward it to the WstethJoin
```
alcueca, line 41:
```suggestion
    /// @dev Wrap stEth held by this Ladle and forward it to the WstEthJoin
```

contracts/modules/LidoModule.sol
```diff
@@ -0,0 +1,58 @@
...
+    /// This function should be called first in a batch, and the Join should keep track of stored reserves
+    /// Passing the id for a join that doesn't link to a contract implemnting IWETH9 will fail
```
alcueca, line 43:
```suggestion
    /// Passing the id for a join that doesn't link to a contract implementing IWstETH will fail
```

alcueca (changes requested):
You can have `immutable` state variables in a module, and assign them as usual in the constructor.

I've suggested how to do that, instead of reusing the `weth` variable from LadleStorage, and done a bit of renaming

alcueca:
I've realized that we don't need this to be a module at all.

It can be a standalone contract, with two functions:
`wrap(address to)` takes the stEth balance of the contract, wraps it into wstEth, and sends it to `to`.
`unwrap(address to)` takes the wstEth balance of the contract, unwraps it into stEth, and sends it to `to`.

The frontend will take care of replacing `to` with the wstEth join or the user address.

Please refactor in that direction, and test it. You can remove the `LadleStorage` inheritance and the `getJoin` function. You can also remove most of the imports and `using` statements.

contracts/modules/LidoWrapHandler.sol
```diff
@@ -0,0 +1,46 @@
...
+
+import '@yield-protocol/utils-v2/contracts/token/IERC20.sol';
+import '@yield-protocol/utils-v2/contracts/token/TransferHelper.sol';
+import '../LadleStorage.sol';
+import '../oracles/lido/IWstETH.sol';
+
+/// @dev A contract to handle wrapping & unwrapping of stETH
+contract LidoWrapHandler {
+    using TransferHelper for IERC20;
+    using TransferHelper for IWstETH;
```
alcueca, line 12: Does this work? I think that `TransferHelper.safeTransfer()` takes an IERC20 as the first parameter.

When using `using`, the function can be appended to a variable with the same type as its first parameter.

Below, you can cast wstETH into an IERC20 by doing `IERC20(address(wstETH))`

contracts/modules/LidoWrapHandler.sol
```diff
@@ -0,0 +1,46 @@
...
+    constructor(IWstETH wstETH_, IERC20 stETH_) {
+        wstETH = wstETH_;
+        stETH = stETH_;
+    }
+
+    /// @dev Wrap stEth held by this contract and forward it to the 'to' address
+    function wrap(address to) external returns (uint256 wrappedAmount) {
+        uint256 stEthTransferred = stETH.balanceOf(address(this));
+        require(stEthTransferred > 0, 'No stETH to wrap');
+        stETH.approve(address(wstETH), stEthTransferred);
```
alcueca, line 32: You can do this in the constructor and save some gas for each user.

`stEth_.approve(address(wstEth_), type(uint256).max)`

It is ok because this contract is not expected to hold any tokens between transactions, to that max approve can't be used to steal anything.

contracts/modules/LidoWrapHandler.sol
```diff
@@ -0,0 +1,46 @@
...
+        wrappedAmount = wstETH.wrap(stEthTransferred);
+        wstETH.safeTransfer(to, wrappedAmount);
+        emit Wrapped(to, wrappedAmount);
+    }
+
+    /// @dev Unwrap WstETH held by this contract, and send the stETH to the 'to' address
+    function unwrap(address to) external returns (uint256 unwrappedAmount) {
+        uint256 wstEthTransferred = wstETH.balanceOf(address(this));
+        require(wstEthTransferred > 0, 'No wstETH to unwrap');
+        unwrappedAmount = wstETH.unwrap(wstEthTransferred);
```
alcueca, line 42: Does the real WstETH contract expect the user to have approved the wstEth tokens to be taken by the WstETH contract for unwrapping?

test/091_lido_wrapperunwrapper.ts
```diff
@@ -0,0 +1,82 @@
...
+import { constants, id } from '@yield-protocol/utils-v2'
+const { WAD } = constants
+import { ETH, DAI, USDC, WSTETH, STETH } from '../src/constants'
+
+import { ethers, waffle } from 'hardhat'
+import { expect } from 'chai'
+const { deployContract } = waffle
+
+import { STETHMock } from '../typechain/STETHMock'
+import { WstETHMock } from '../typechain/WSTETHMock'
```
alcueca, line 11: If the tests pass for you locally, try removing the `artifacts`, `typechain` and `cache` folders, and then doing `yarn build` and `yarn test` again to replicate this error. I can also try.

alcueca (changes requested):
Getting there, one suggestion about using TransferHelper and libraries in general, one gas optimization, one question about approves, and the tests that fail in CI.

contracts/modules/LidoWrapHandler.sol
iamsahu: The real contract doesn't expect an approval. Here is the code for unwrap: https://etherscan.io/address/0x7f39c581f595b53c5cb19bd0b3f8da6c935e2ca0#code#L1090

test/091_lido_wrapperunwrapper.ts
iamsahu: I see the mistake now. Have reverted to using ERC20Mock for steth.

contracts/modules/LidoWrapHandler.sol
iamsahu: This is working in the local environment. IWstETH is inheriting from IERC20 that might be the reason for this to work?

alcueca: Ah, that's the reason, cool!

contracts/modules/LidoWrapHandler.sol
alcueca: :+1:

alcueca (approved):
Looks good to go, good job :)

PR #312, opened 2021-11-05
feat: YearnVaultMultiOracle

richiehumphrey:
This pr adds a new multi oracle for checking prices of yearn vault tokens.  

 - The pricing mechanism is based on the yearn vault token `pricePerShare()` function which is [calculated by dividing the total assets by the total supply of tokens](https://github.com/yearn/yearn-vaults/blob/2a28331438063a48684db4143ad293821671cc67/contracts/Vault.vy#L1141).  

 - When setting source, the inverted pair is also set

```
  Oracles - Yearn Vault
    ✓ get() reverts if pair not found
    ✓ setSource() reverts if quote and base are the same
    ✓ setSource() sets a pair and the inverse pair
    with sources set
      ✓ get() and peek() return correct values
      ✓ get() reverts on zero price 

```

```
·---------------------------------------|---------------------------|--------------|-----------------------------·
|          Solc version: 0.8.6          ·  Optimizer enabled: true  ·  Runs: 1000  ·  Block limit: 12450000 gas  │
········································|···························|··············|······························
|  Methods                                                                                                       │
··························|·············|·············|·············|··············|···············|··············
|  Contract               ·  Method     ·  Min        ·  Max        ·  Avg         ·  # calls      ·  usd (avg)  │
··························|·············|·············|·············|··············|···············|··············
|  YearnVaultMock         ·  setPrice   ·      13225  ·      26485  ·       19855  ·            2  ·          -  │
··························|·············|·············|·············|··············|···············|··············
|  YearnVaultMultiOracle  ·  grantRole  ·          -  ·          -  ·       51115  ·            1  ·          -  │
··························|·············|·············|·············|··············|···············|··············
|  YearnVaultMultiOracle  ·  setSource  ·      40440  ·      80240  ·       60340  ·            6  ·          -  │
··························|·············|·············|·············|··············|···············|··············
|  Deployments                          ·                                          ·  % of limit   ·             │
········································|·············|·············|··············|···············|··············
|  USDCMock                             ·          -  ·          -  ·      847817  ·        6.8 %  ·          -  │
········································|·············|·············|··············|···············|··············
|  YearnVaultMock                       ·     907804  ·     907816  ·      907810  ·        7.3 %  ·          -  │
········································|·············|·············|··············|···············|··············
|  YearnVaultMultiOracle                ·          -  ·          -  ·     1096162  ·        8.8 %  ·          -  │
·---------------------------------------|-------------|-------------|--------------|---------------|-------------·


```


iamsahu (approved):

contracts/oracles/yearn/YearnVaultMultiOracle.sol
```diff
@@ -0,0 +1,140 @@
...
+
+    /**
+     *@notice Set or reset an oracle source and its inverse
+     *@param  baseId id used for related base token (e.g. USDC)
+     *@param  vaultTokenId id for Yearn vault token (e.g. yvUSDC)
+     *@param  vaultToken address for Yearn vault token
+     */
+    function setSource(
+        bytes6 baseId,
+        bytes6 vaultTokenId,
```
alcueca, line 48: Is it necessary to require the yvToken to be a specific parameter? This could be `quoteId` and if you pass the yvToken as the first parameter everything would still work, right?

contracts/oracles/yearn/IYvToken.sol
```diff
@@ -0,0 +1,11 @@
+// SPDX-License-Identifier: MIT
+pragma solidity >=0.6.0;
+import "@yield-protocol/utils-v2/contracts/token/IERC20.sol";
+
+// @notice Interface for Yearn Vault tokens for use with Yield price oracles
+// @dev see https://github.com/yearn/yearn-vaults/blob/main/contracts/Vault.vy
+interface IYvToken is IERC20 {
```
alcueca, line 7: I think this needs to inherit from IERC20Metadata.

alcueca (changes requested):
Looks good to me, but please just look at the one comment, maybe we can make governance around this a bit more robust.

contracts/oracles/yearn/YearnVaultMultiOracle.sol
richiehumphrey: yes good call

src/constants.ts
```diff
@@ -2,8 +2,15 @@ import { ethers } from 'ethers'
...
 export const RATE = ethers.utils.formatBytes32String('RATE').slice(0, 14)
+
 export const DAI = ethers.utils.formatBytes32String('01').slice(0, 14)
 export const USDC = ethers.utils.formatBytes32String('02').slice(0, 14)
 export const ETH = ethers.utils.formatBytes32String('00').slice(0, 14)
 export const STETH = ethers.utils.formatBytes32String('05').slice(0, 14)
 export const WSTETH = ethers.utils.formatBytes32String('06').slice(0, 14)
+
+//yearn vaults 1xx
+export const YVETH = ethers.utils.formatBytes32String('100').slice(0, 14)
```
richiehumphrey, line 13: Hmm, first of all I just realized we are keeping this list id's in (at least) 2 places -- the other one is in the [environments repo](https://github.com/yieldprotocol/environments-v2/blob/main/shared/constants.ts)  How would you like to handle this?  I don't think it's a big deal.  We can just leave it for now, maybe I'll add a comment to make sure we remember to update both places?  Alternatively we can move these constants to yield-utils or somewhere and have that be the source of truth

src/constants.ts
```diff
@@ -2,8 +2,15 @@ import { ethers } from 'ethers'
...
+export const YVDAI = ethers.utils.formatBytes32String('101').slice(0, 14)
```
richiehumphrey, line 14: second question, just confirming, would you like me to change these yearn vault constants to the next consecutive ordinal numbers as you suggested in a recent code review comment?

If I did that, I would not include yvSTETH or yvETH in this pr.  and make yvDai and yvUsdc be "08" and "09" after ens

I would add yvUSDC and yvDai as "09" and "10" respectively

src/constants.ts
alcueca: Let's march environments for now, that's where the master list should live imho.

src/constants.ts
alcueca: Yes please, let's stick to consecutive ordinals.

alcueca (approved):
Let's ship toot toot :ship:

PR #313, opened 2021-11-16
tweak: removed setSecondsAgo function in UniswapV3 Oracle

iamsahu:
# Goal
Remove the setSecondsAgo as a separate function

# Details
1. Removed the function to setSecondsAgo and moved setting it to the setSource function.
2. Source data will store the secondsAgo value for each source
3. Modified test to reflect the changes in the setSource function 

# Test
Run [015_oracles_uniswap](https://github.com/yieldprotocol/vault-v2/blob/tweak/uniswapv3-oracle/test/015_oracles_uniswap.ts)

### DIFF
--- a/contracts/oracles/uniswap/UniswapV3Oracle.sol
+++ b/contracts/oracles/uniswap/UniswapV3Oracle.sol
@@ -1,13 +1,13 @@
 // SPDX-License-Identifier: BUSL-1.1
 pragma solidity 0.8.6;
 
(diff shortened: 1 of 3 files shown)


alcueca (changes requested):
I'm ok with the change, but I don't think that you should have passed the linter, it introduces a lot of noise in the diff.

We can't really pass a linter on the code that has been audited, so I don't think we should use a linter at all in the smart contracts, tbh.

Could you please revert the linting changes?

alcueca (approved):
Nice and clean, thank you

PR #317, opened 2021-11-18
AccumulatorMultiOracle: accumulator-based multi oracle

sblOWPCKCR:


contracts/oracles/accumulator/AccumulatorMultiOracle.sol
```diff
@@ -0,0 +1,128 @@
...
+     */
+    function get(
+        bytes32 base,
+        bytes32 kind,
+        uint256
+    ) external virtual override returns (uint256 accumulated, uint256 updateTime) {
+        Accumulator memory accumulator = sources[base.b6()][kind.b6()];
+        require(accumulator.accumulated != 0, "Source not found");
+
+        uint256 cycles = (block.timestamp - accumulator.lastUpdated);
```
alcueca, line 116: Take into account that there could be several requests in each block, and with the current implementation we would waste gas.

If `cycles == 0` you should skip the updating, and just return the stored value.

Also, `cycles` is a strange choice for the variable, `secondsAgo`, `secondsSinceUpdated`?

alcueca (changes requested):
I did some general renaming, and left one comment about saving gas when there is more than one request per block, otherwise all good.

alcueca (approved):
Good to go!

PR #318, opened 2021-11-18
Use right imports in UniswapV3Oracle and gas golf

alcueca:
We were still using mocks in the UniswapV3Oracle (unlike in @iamsahu's repo).

I corrected the imports to use the solc 0.8 uniswap libraries in @iamsahu's repo, and went in to fix a few more things:
 - Renaming of variables (source, value, amount, secondsAgo)
 - Removal of unused data (factory, fee)
 - Merging of Source and SourceData. They just fit in two words, vs three when separate (and with factory and fee). It uses a bit more gas when setting a new source, but it saves 1 SLOAD when reading the oracle. Also clearer code.
 - Found and fixed an unchecked cast from uint256 to uint128
 - Exported via npm

I've tested this with the ENS test in environments-v2, and works. I couldn't get the test from @iamsahu to work, though. I wonder if it is an issue with the ABI.


Eihcir0 (approved):

Eihcir0 (approved):

richiehumphrey (approved):

iamsahu (approved):
Looks good to go. It is much easier to read now after the removal of the source.inverse

alcueca:
I'm finding it quite hard to make the tests work, due to dependencies. I'm closing this branch and putting the changes in a [fresh one](https://github.com/yieldprotocol/vault-v2/pull/324), with no package.json updates.

PR #320, opened 2021-11-19
ChainlinkUSDMultiOracle: A -> B via ChainLink USD oracle

sblOWPCKCR:


contracts/oracles/chainlink/ChainlinkUSDMultiOracle.sol
```diff
@@ -0,0 +1,110 @@
...
+        amountQuote = (amountBase * basePrice) / quotePrice;
+
+        // now need to convert baseDecimals to quoteDecimals
+        if (baseDecimals <= quoteDecimals) {
+            amountQuote *= (10**(quoteDecimals - baseDecimals));
+        } else {
+            amountQuote /= (10**(baseDecimals - quoteDecimals));
+        }
+
+        if (updateTime1 < updateTime2) {
```
alcueca, line 104:
```suggestion
        updateTime = (updateTime1 < updateTime2) ? updateTime1 : updateTime2;
```

test/018_oracles_chainlink_usd.ts
```diff
@@ -0,0 +1,120 @@
...
+
+  it('does not allow non-8-digit USD sources', async () => {
+    // base and quote are missing
+    await expect(
+      oracle.setSource(baseId1, tokenA.address, (await genChainlinkAggregatorMock(18, BigNumber.from(1))).address)
+    ).to.be.revertedWith('Non-8-decimals USD source')
+  })
+
+  // BigNumber doesn't support real numbers -> we use WADs
+  for (const [a_in_usd, b_in_usd, expected_b_for_1_a_in_WAD] of [
```
alcueca, line 85: I didn't know you could put tests inside loops. Neat.

alcueca (approved):
There is a minor thing about using a ternary operator for compactness, but that's it.

I thought for quite a long time about your implementation vs the ChainlinkMultiOracle one. While I like the `_peek` and `_peekThroughETH` implementations better, maybe because they are a bit more generalistic, but I can't say that your implementation doesn't fit exactly the use case.

Using the ChainlinkMultiOracle implementation would make you deal with inverses, just to be able to obtain prices in USD (and possibly manage decimals in a more robust manner), but I don't think it's worth it.

Well done.

contracts/oracles/chainlink/ChainlinkUSDMultiOracle.sol
sblOWPCKCR: what does the actual logic should be? my original code returns the bigger of the two times, your returns the smaller

sblOWPCKCR:
the biggest problem with using ChainlinkMultiOracle is that USD doesn't have a token, we'd have to create a fake one

contracts/oracles/chainlink/ChainlinkUSDMultiOracle.sol
sblOWPCKCR: nvm, your version is correct

alcueca (approved):
Good to go 👍

PR #322, opened 2021-11-19
fix: setSource params

richiehumphrey:
Ok here is (hopefully) final version with all comments addressed

```
  Oracles - Yearn Vault
    ✓ get() reverts if pair not found
    ✓ setSource() sets a pair and the inverse pair
    with sources set
      ✓ get() and peek() return correct values
      ✓ get() reverts on zero price 

·---------------------------------------|---------------------------|--------------|-----------------------------·
|          Solc version: 0.8.6          ·  Optimizer enabled: true  ·  Runs: 1000  ·  Block limit: 12450000 gas  │
········································|···························|··············|······························
|  Methods                                                                                                       │
··························|·············|·············|·············|··············|···············|··············
|  Contract               ·  Method     ·  Min        ·  Max        ·  Avg         ·  # calls      ·  usd (avg)  │
··························|·············|·············|·············|··············|···············|··············
|  YearnVaultMock         ·  setPrice   ·      13225  ·      26485  ·       19855  ·            2  ·          -  │
··························|·············|·············|·············|··············|···············|··············
|  YearnVaultMultiOracle  ·  grantRole  ·          -  ·          -  ·       51115  ·            1  ·          -  │
··························|·············|·············|·············|··············|···············|··············
|  YearnVaultMultiOracle  ·  setSource  ·      39809  ·      79609  ·       59709  ·            6  ·          -  │
··························|·············|·············|·············|··············|···············|··············
|  Deployments                          ·                                          ·  % of limit   ·             │
········································|·············|·············|··············|···············|··············
|  USDCMock                             ·          -  ·          -  ·      847817  ·        6.8 %  ·          -  │
········································|·············|·············|··············|···············|··············
|  YearnVaultMock                       ·     907804  ·     907816  ·      907810  ·        7.3 %  ·          -  │
········································|·············|·············|··············|···············|··············
|  YearnVaultMultiOracle                ·          -  ·          -  ·     1074254  ·        8.6 %  ·          -  │
·---------------------------------------|-------------|-------------|--------------|---------------|-------------·

  4 passing (3s)

```


contracts/oracles/yearn/YearnVaultMultiOracle.sol
```diff
@@ -37,22 +37,19 @@ contract YearnVaultMultiOracle is IOracle, AccessControl {
...
      */
     function setSource(
-        bytes6 baseId,
-        bytes6 quoteId,
+        bytes6 underlyingId,
         IYvToken vaultToken
     ) external auth {
-        require(baseId != quoteId, "Set source not needed");
-
+        bytes6 vaultTokenId = bytes6(bytes(vaultToken.symbol()));
```
alcueca, line 48: The `vaultTokenId` needs to be passed as a parameter (there isn't a guarantee that the yvToken symbol will fit on 6 bytes, or that it won't clash with something else)

src/constants.ts
```diff
@@ -11,8 +11,8 @@ export const WSTETH = ethers.utils.formatBytes32String('04').slice(0, 14)
 export const STETH = ethers.utils.formatBytes32String('05').slice(0, 14)
 export const LINK = ethers.utils.formatBytes32String('06').slice(0, 14)
 export const ENS = ethers.utils.formatBytes32String('07').slice(0, 14)
-export const YVDAI = ethers.utils.formatBytes32String('08').slice(0, 14)
-export const YVUSDC = ethers.utils.formatBytes32String('09').slice(0, 14)
+export const YVDAI = ethers.utils.formatBytes32String('yvDAI').slice(0, 14)
```
alcueca, line 14: We are following the sequential ordering of asset ids, with the asset id determined at the time that the proposal is submitted, please choose '08' and '09' here.

Those ids might change by the time we get to mainnet, but that change would be reflected only in the environments repo. The source of truth for asset ids will always be the Cauldron, and you can recover all asset ids either by filtering the events, or querying the Cauldron for sequentially increasing asset ids until you get one that gives you the zero address.

alcueca (changes requested):
Sorry, but the yearnVaultId must be chosen externally, and passed on as a parameter. We are also sticking with the sequential ordering.

alcueca (changes requested):
Good to go!

alcueca (approved):
Oops, wrong button. Good to go, I meant.

PR #326, opened 2021-11-22
ChainlinkL2USDMultiOracle: Chainlink USD Oracle that's safe to use on L2

sblOWPCKCR:
Incorporates https://docs.chain.link/docs/l2-sequencer-flag/

Instead of parametrizing ChainlinkUSDMultiOracle, we create new contract to avoid
increasing the gas cost of ChainlinkUSDMultiOracle when there's no need to check the
sequencer flag


contracts/oracles/chainlink/ChainlinkUSDMultiOracle.sol
```diff
@@ -104,3 +105,57 @@ contract ChainlinkUSDMultiOracle is IOracle, AccessControl, Constants {
...
+    // https://docs.chain.link/docs/l2-sequencer-flag/
+    address internal constant FLAG_ARBITRUM_SEQ_OFFLINE =
+        address(bytes20(bytes32(uint256(keccak256('chainlink.flags.arbitrum-seq-offline')) - 1)));
+
+    constructor(FlagsInterface flags) {
+        require(address(flags) != address(0), 'FlagsInterface has to be set');
+        chainlinkFlags = flags;
+    }
+
+    function assertCanGetPrice() private view {
```
alcueca, line 129: This looks like a good candidate for a modifier:
```
modifier onlyFresh() {
  // If flag is raised we shouldn't perform any critical operations
  require(
    chainlinkFlags.getFlag(FLAG_ARBITRUM_SEQ_OFFLINE) != isRaised,
    "Chainlink feeds are not being updated"
  );
  _;
}
```

contracts/oracles/chainlink/ChainlinkUSDMultiOracle.sol
```diff
@@ -104,3 +105,57 @@ contract ChainlinkUSDMultiOracle is IOracle, AccessControl, Constants {
...
+    function peek(
+        bytes32 baseId,
+        bytes32 quoteId,
+        uint256 amountBase
+    ) external view virtual override returns (uint256 amountQuote, uint256 updateTime) {
+        assertCanGetPrice();
+        if (baseId == quoteId) return (amountBase, block.timestamp);
+
+        (amountQuote, updateTime) = _peekThroughUSD(baseId.b6(), quoteId.b6(), amountBase);
+    }
```
alcueca, line 147: Simpler code:
```
    /// @dev Convert amountBase base into quote at the latest oracle price.
    function peek(bytes32 baseId, bytes32 quoteId, uint256 amountBase)
      external view virtual override
      onlyFresh
      returns (uint256 amountQuote, uint256 updateTime)
    {
      return super.peek(baseId, quoteId, amountBase);
    }
```

contracts/oracles/chainlink/ChainlinkUSDMultiOracle.sol
```diff
@@ -104,3 +105,57 @@ contract ChainlinkUSDMultiOracle is IOracle, AccessControl, Constants {
...
+        bytes32 quoteId,
+        uint256 amountBase
+    ) external virtual override returns (uint256 amountQuote, uint256 updateTime) {
+        assertCanGetPrice();
+
+        if (baseId == quoteId) return (amountBase, block.timestamp);
+
+        (amountQuote, updateTime) = _peekThroughUSD(baseId.b6(), quoteId.b6(), amountBase);
+    }
+}
```
alcueca, line 161: As before:
```
    function get(bytes32 baseId, bytes32 quoteId, uint256 amountBase)
      external virtual override
      onlyFresh
      returns (uint256 amountQuote, uint256 updateTime)
    {
      return super.get(baseId, quoteId, amountBase);
    }
```

alcueca (changes requested):
Fine in principle, but the code can be simpler. Interesting stuff.

contracts/oracles/chainlink/ChainlinkUSDMultiOracle.sol
sblOWPCKCR: nope, can't do that for `external` functions

contracts/oracles/chainlink/ChainlinkUSDMultiOracle.sol
sblOWPCKCR: nope, can't do that for `external` functions

contracts/oracles/chainlink/ChainlinkUSDMultiOracle.sol
sblOWPCKCR: I find modifiers easy to miss, so went for an explicit check in the function body
Also, remember Tenderly always pointing you at `auth` modifier as the source of the revert? That's another problem with modifiers

With that said, I'm not super opposed to them. Took your suggestion

contracts/oracles/chainlink/ChainlinkUSDMultiOracle.sol
```diff
@@ -104,3 +105,52 @@ contract ChainlinkUSDMultiOracle is IOracle, AccessControl, Constants {
...
 }
+
+/**
+ * @title ChainlinkL2USDMultiOracle: ChainlinkUSDMultiOracle that's safe to use on L2
+ * @notice Chainlink recommends checking the sequencer status on some L2 networks to avoid
+  reading stale data
+
+  https://docs.chain.link/docs/l2-sequencer-flag/
+ */
+contract ChainlinkL2USDMultiOracle is ChainlinkUSDMultiOracle {
```
alcueca, line 116: Didn't notice this before, could you please split ChainlinkL2USDMultiOracle into its own file?

alcueca (changes requested):
One contract per file, please.

contracts/oracles/chainlink/ChainlinkUSDMultiOracle.sol
alcueca: Oh well

PR #328, opened 2021-11-24
Convex as a collateral

iamsahu:
# Goal
Use convex tokens as a collateral


contracts/utils/convex/ConvexStakingWrapperYield.sol
```diff
@@ -0,0 +1,111 @@
...
+
+    function initialize(
+        address _curveToken,
+        address _convexToken,
+        address _convexPool,
+        uint256 _poolId,
+        address _vault
+    ) external override auth{
+        require(!isInit, 'already init');
+        owner = address(0xa3C5A1e09150B75ff251c1a7815A07182c3de2FB); //TODO: Find why this needs to be set to convex multisig
```
alcueca, line 39: This would be our Timelock, instead of the convex multisig.

contracts/utils/convex/ConvexStakingWrapperYield.sol
```diff
@@ -0,0 +1,111 @@
...
+        emit OwnershipTransferred(address(0), owner);//TODO: Find why this needs to be done
```
alcueca, line 40: This is done so that there is an event trail for ownership changes, including the first one

contracts/utils/convex/ConvexStakingWrapperYield.sol
```diff
@@ -0,0 +1,111 @@
...
+        convexPool = _convexPool;
+        convexPoolId = _poolId;
+        collateralVault = _vault; //TODO: Add the join address
+
+        //add rewards
+        addRewards();
+        setApprovals();
+    }
+
+    function setCauldron(address _cauldron) external auth{
```
alcueca, line 56: This should be a parameter to `initialize`

contracts/utils/convex/ConvexStakingWrapperYield.sol
```diff
@@ -0,0 +1,111 @@
...
+    function setVault(address _account, bytes12 _vault) external auth {
+        bytes12[] storage userVault = vaults[_account];
+        for (uint256 i = 0; i < userVault.length; i++) {
+            require(userVault[i] != _vault, 'already added');
+        }
+        userVault.push(_vault);
+        vaults[_account] = userVault;
+    }
+
+    function removeVault(address _account, bytes12 _vault) external auth {
```
alcueca, line 71: It will be easier to never remove vaults

contracts/utils/convex/ConvexStakingWrapperYield.sol
```diff
@@ -0,0 +1,111 @@
...
+                vaults[_account] = remove(i,userVault);
+                break;
+            }
+        }
+    }
+
+    function remove(uint _index,bytes12[] storage userVault) internal returns (bytes12[] memory){
+        require(_index < userVault.length, "index out of bound");
+
+        for (uint i = _index; i < userVault.length - 1; i++) {
```
alcueca, line 84: Instead of this very expensive loop (1 SSTORE for position after _index), if `_index` is not the last position in the array, copy the value of the last position to `_index` and then `pop`. You lose the ordering, but we don't care about that.

contracts/utils/convex/ConvexStakingWrapperYield.sol
```diff
@@ -0,0 +1,111 @@
...
+        }
+
+        if (vaults[_account].length == 0) {
+            return balanceOf(_account);
+        }
+    bytes12[] memory userVault = vaults[_account];
+        //add up all balances of all vaults
+        uint256 collateral;
+        for (uint256 i = 0; i < userVault.length; i++) {
+            try ICauldron(cauldron).balances(userVault[i]) returns (Balances memory _balance) {
```
alcueca, line 104: No need to use `try... catch`, if you read `cauldron.balances` on a non existing `vaultId` you will get an empty `Balances` struct, with `ink` and `art` set to zero.

So you can just assume the balances exist and add them.

contracts/utils/convex/ConvexStakingWrapperYield.sol
```diff
@@ -0,0 +1,111 @@
...
+        setApprovals();
+    }
+
+    function setCauldron(address _cauldron) external auth{
+        require(_cauldron!=address(0), 'cauldron address cannot be 0');
+        cauldron = _cauldron;
+    }
+
+    // Set the locations of vaults where the user's funds have been deposited & the accounting is kept
+    function setVault(address _account, bytes12 _vault) external auth {
```
alcueca, line 62: Instead of an `auth` function, you can remove the `account` parameter, and read the owner from the cauldron using `cauldron.vaults(vaultId).owner`. Anyone can register anyone's vault, which is fine.

Then, we can have a Ladle module that takes a vaultId as parameter and calls this function, so that we can embed `setVault` in a Ladle batch.

It would be a module, because the Ladle keeps in `cachedVaultId` the vaultId of a vault that would just have been created in the same batch. If `convexModule.setVault` receives a zero as the vaultId, it calls `convexStakingWrapperYield.setVault(cachedVaultId)`, otherwise it calls `convexStakingWrapperYield.setVault(vaultId)`.

contracts/utils/convex/ConvexStakingWrapperYield.sol
```diff
@@ -0,0 +1,111 @@
...
+    using SafeERC20 for IERC20;
+    using Address for address;
+
+
+    mapping(address=>bytes12[]) public vaults;
+    address cauldron;
+
+    constructor() public {}
+
+    function initialize(
```
alcueca, line 31: I would just add all this into the constructor instead.

contracts/utils/convex/ConvexStakingWrapperYield.sol
```diff
@@ -0,0 +1,111 @@
...
+}
+
+interface ICauldron {
+    /// @dev Each vault records debt and collateral balances_.
+    function balances(bytes12 vault) external view returns (Balances memory);
+}
+
+//Staking wrapper for Yield platform
+//use convex LP positions as collateral while still receiving rewards
+contract ConvexStakingWrapperYield is ConvexStakingWrapper,AccessControl {
```
alcueca, line 21: You don't actually need `AccessControl`. The parent class uses `Ownable`, which we could reuse if we would need to. If you keep reading you'll see we need no permissioned functions.

alcueca (changes requested):
In the comments you'll find out how to simplify the contract, and integrate it in the Ladle through a very simple module.

Please have a go through your code as well and fix the spacing (both sides of an  `=>`, before a `{`, both side of a `!=`).

contracts/utils/convex/ConvexStakingWrapperYield.sol
iamsahu: Should we also handle the wrapping & unwrapping of the convex token via this module? This would allow the user to deposit the convex token and we could take care of the wrapping & unwrapping for them.

iamsahu: My last comment could be ignored as we can directly add the wrapper as an integration and handle it through that.

contracts/utils/convex/ConvexStakingWrapperYield.sol
```diff
@@ -0,0 +1,119 @@
...
+        convexPoolId = poolId_;
+        collateralVault = join_; //TODO: Add the join address
+        cauldron = cauldron_;
+
+        //add rewards
+        addRewards();
+        setApprovals();
+    }
+
+    function setCollateralVault(address join_) external {
```
alcueca, line 52: Please make this `auth` instead, and emit an event at the end.

Please also rename this function as `point`. We use that as the function name every time time that we allow to change the orchestration after deployment.

contracts/utils/convex/ConvexStakingWrapperYield.sol
```diff
@@ -0,0 +1,119 @@
...
+        require(msg.sender==owner,'Only owner can set vault');
+        collateralVault = join_;
+    }
+
+    /// @notice Adds a vault to the user's vault list
+    /// @param vault_ The vaulId being added
+    function addVault(bytes12 vault_) external {
```
alcueca, line 59: Rename `vault_` as `vaultId`

contracts/utils/convex/ConvexStakingWrapperYield.sol
```diff
@@ -0,0 +1,119 @@
...
+        address account = cauldron.vaults(vault_).owner;
+        require(account != address(0), 'No owner for the vault');
+        bytes12[] storage userVault = vaults[account];
```
alcueca, line 62: Rename `userVault` as `vaults_`

contracts/utils/convex/ConvexStakingWrapperYield.sol
```diff
@@ -0,0 +1,119 @@
...
+        userVault.push(vault_);
+        vaults[account] = userVault;
+        emit VaultSet(account, vault_);
+    }
+
+    /// @notice Get user's balance of collateral deposited in various vaults
+    /// @param account_ User's address for which balance is requested
+    /// @return User's balance of collateral
+    function _getDepositedBalance(address account_) internal view override returns (uint256) {
+        if (account_ == address(0) || account_ == collateralVault) {
```
alcueca, line 75: Why do we return zero for `address(0)`?

contracts/utils/convex/ConvexStakingWrapperYield.sol
```diff
@@ -0,0 +1,119 @@
...
+        for (uint256 i = 0; i < userVault.length; i++) {
+            balance = cauldron.balances(userVault[i]);
+            collateral = collateral + balance.ink;
+        }
+
+        //add to balance of this token
+        return _balanceOf[account_] + collateral;
+    }
+
+    function withdrawFor(uint256 _amount,address _account) external nonReentrant {
```
alcueca, line 96: Natspec, please

contracts/utils/convex/ConvexStakingWrapperYield.sol
```diff
@@ -0,0 +1,119 @@
...
+        if (_amount > 0) {
+            _burn(_account, _amount);
+            IRewardStaking(convexPool).withdraw(_amount, false);
+            IERC20(convexToken).safeTransfer(_account, _amount);
+        }
+
+        emit Withdrawn(_account, _amount, false);
+    }
+
+    function stakeFor(uint256 _amount,address account, address _to) external nonReentrant {
```
alcueca, line 107: Natspec, please

alcueca (changes requested):
I think we can refactor the original contracts so that they are in tune with our codebase.

contracts/utils/convex/ConvexStakingWrapper.sol
```diff
@@ -0,0 +1,452 @@
...
+    bool public isInit;
+    address public owner;
+
+    string internal _tokenname;
+    string internal _tokensymbol;
+
+    uint256 private constant _NOT_ENTERED = 1;
+    uint256 private constant _ENTERED = 2;
+
+    uint256 private _status;
```
sblOWPCKCR, line 88: minor: _status could be `bool` and moved up after line 79 - I believe this will lead to better storage packing (fewer slots used)

contracts/utils/convex/ConvexStakingWrapper.sol
```diff
@@ -0,0 +1,452 @@
...
+        owner = address(0);
+    }
+
+    function shutdown() external onlyOwner {
+        isShutdown = true;
+    }
+
+    function setApprovals() public {
+        IERC20(curveToken).approve(convexBooster, 0);
+        IERC20(curveToken).approve(convexBooster, type(uint128).max);
```
sblOWPCKCR, line 170: type(uint256).max? here and at line 172

contracts/utils/convex/ConvexStakingWrapper.sol
```diff
@@ -0,0 +1,452 @@
...
+    ) internal {
+        RewardType storage reward = rewards[_index];
+
+        //get difference in balance and remaining rewards
+        //getReward is unguarded so we use reward_remaining to keep track of how much was actually claimed
+        uint256 bal = IERC20(reward.reward_token).balanceOf(address(this));
+        // uint256 d_reward = bal-(reward.reward_remaining);
+
+        if (_supply > 0 && bal - (reward.reward_remaining) > 0) {
+            reward.reward_integral = reward.reward_integral + uint128(bal - (reward.reward_remaining * 1e20) / _supply);
```
sblOWPCKCR, line 272: the original code is:
```
reward.reward_integral = reward.reward_integral + uint128(bal.sub(reward.reward_remaining).mul(1e20).div(_supply));
```
Which is: `reward_integral = reward_integral + (bal - reward_remaining) * 1e20 / _supply`
You're doing `reward_integral = reward_integral + (bal - reward_remaining * 1e20 / _supply)`

sblOWPCKCR (dismissed):

contracts/utils/convex/ConvexStakingWrapper.sol
```diff
@@ -0,0 +1,452 @@
...
+    function renounceOwnership() public virtual onlyOwner {
+        emit OwnershipTransferred(owner, address(0));
+        owner = address(0);
+    }
+
+    function shutdown() external onlyOwner {
+        isShutdown = true;
+    }
+
+    function setApprovals() public {
```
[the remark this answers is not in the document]
iamsahu: They must be going for the clonefactory pattern which is usually supplemented with the initialize function.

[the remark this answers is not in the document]
iamsahu: They must be going for the clone factory pattern. You can see this function being called in the initialize function.

contracts/utils/convex/ConvexStakingWrapperYield.sol
```diff
@@ -0,0 +1,119 @@
...
+
+    constructor(
+        address curveToken_,
+        address convexToken_,
+        address convexPool_,
+        uint256 poolId_,
+        address join_,
+        ICauldron cauldron_,
+        address timelock_
+    ) {
```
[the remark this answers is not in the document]
iamsahu: @alcueca Can you please give me an example of this, I have not been able to understand what you mean by this.

alcueca:
What's the status with this, @iamsahu? Would you like me to review it again?

contracts/utils/convex/ConvexStakingWrapperYield.sol
```diff
@@ -0,0 +1,112 @@
...
+
+    /// @notice Get user's balance of collateral deposited in various vaults
+    /// @param account_ User's address for which balance is requested
+    /// @return User's balance of collateral
+    function _getDepositedBalance(address account_) internal view override returns (uint256) {
+        if (account_ == address(0) || account_ == collateralVault) {
+            return 0;
+        }
+
+        if (vaults[account_].length == 0) {
```
alcueca, line 72: Someone can own tokens of this wrapper, plus also all the tokens that has posted as collateral in Yield :+1:

contracts/utils/convex/ConvexStakingWrapperYield.sol
```diff
@@ -0,0 +1,112 @@
...
+            collateral = collateral + balance.ink;
+        }
+
+        //add to balance of this token
+        return _balanceOf[account_] + collateral;
+    }
+
+    /// @dev Wrap cvx3CRV held by this contract and forward it to the 'to' address
+    /// @param to_ Address to send the wrapped token to
+    function wrap(address to_, address from_) external {
```
alcueca, line 91: Who is `from_` in this case?

contracts/utils/convex/ConvexStakingWrapperYield.sol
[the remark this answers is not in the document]
alcueca: Example give on slack

contracts/utils/convex/ConvexStakingWrapperYield.sol
iamsahu: The user whose token is being wrapped. This parameter was added to remove the usage of `tx.origin`.
Will add the natspec

contracts/utils/convex/ConvexStakingWrapperYield.sol
```diff
@@ -0,0 +1,112 @@
...
+
+    /// @notice Points the collateral vault to the join storing the wrappedCvx3Crv
+    /// @param join_ Join which will store the wrappedCvx3Crv of the user
+    function point(address join_) external auth {
+        collateralVault = join_;
+    }
+
+    /// @notice Adds a vault to the user's vault list
+    /// @param vaultId The vaulId being added
+    function addVault(bytes12 vaultId) external {
```
alcueca, line 52: We need to add a function that also removes a vault, since vaults can be given to another user

contracts/utils/convex/ConvexStakingWrapperYield.sol
```diff
@@ -0,0 +1,112 @@
...
+
+        if (vaults[account_].length == 0) {
+            return _balanceOf[account_];
+        }
+        bytes12[] memory userVault = vaults[account_];
+
+        //add up all balances of all vaults
+        uint256 collateral;
+        DataTypes.Balances memory balance;
+        for (uint256 i = 0; i < userVault.length; i++) {
```
alcueca, line 80: Here we need to check that `account_` is still the owner of `userVault[i]`, and call `removeVault` if it is not. We then also need to ignore vaultIds that are zero.

iamsahu: Do we want to do this as it will lead to the getDepositedBalance no longer being a view function which would add some more gas costs everytime user calls the getReward function. Also here are more functions which could no longer be view functions:

1. earned
2. totalBalanceOf

alcueca: `_getDepositedBalance` is an internal function that is only called by mutable (non-view) functions in ConvexYieldWrapper.sol. It makes no difference for it to be `view` or not (`view` functions cost gas when called from mutable functions).

alcueca: The alternative is to leave it as a `view` function, but check the `owner` of each vault before adding its collateral balance. That's an extra `CALL` per vault. It's probably better than making `earned` and `totalBalanceOf` mutable.

contracts/utils/convex/ConvexYieldWrapper.sol
```diff
@@ -61,6 +60,20 @@ contract ConvexStakingWrapperYield is ConvexStakingWrapper {
         emit VaultSet(account, vaultId);
     }
 
+    /// @notice Remove a vault from the user's vault list
+    /// @param vaultId The vaulId being added
+    /// @param account The user from whom the vault needs to be removed
+    function removeVault(bytes12 vaultId, address account) public {
```
alcueca, line 66: To make this permissionless you should pull the `owner` of the vault. If `account != owner` then you remove the vault from `vaults`.

The logic is this:
If a user owns a vault, anyone can register that vault for that owner in the wrapper.
If a user doesn't own a vault, but it is registered for him in the wrapper, anyone can remove it.

Removing vaults is only needed to save a bit of gas on `_getDepositedBalance`, so it doesn't really matter if users sometimes have vaults that don't belong to them. They will be ignored.

contracts/utils/convex/ConvexYieldWrapper.sol
```diff
@@ -69,17 +82,22 @@ contract ConvexStakingWrapperYield is ConvexStakingWrapper {
...
         DataTypes.Balances memory balance;
         for (uint256 i = 0; i < userVault.length; i++) {
-            balance = cauldron.balances(userVault[i]);
-            collateral = collateral + balance.ink;
+            if (userVault[i] != bytes12(0)) {
+                if (cauldron.vaults(userVault[i]).owner == account_) {
+                    balance = cauldron.balances(userVault[i]);
+                    collateral = collateral + balance.ink;
+                }
+                // else {
```
alcueca: I think that adding this call would be fine.

alcueca (commented):
`removeVault` needs to check that the vault owner is different from the account passed on as a parameter

contracts/utils/convex/ConvexYieldWrapper.sol
iamsahu: Adding this would lead to `_getDepositedBalance` no longer being a view only function.

contracts/mocks/ConvexYieldWrapperMock.sol
```diff
@@ -0,0 +1,351 @@
...
+        cauldron = _cauldron;
+    }
+
+    // Set the locations of vaults where the user's funds have been deposited & the accounting is kept
+    function addVault(bytes12 vault_) external {
+        address account = cauldron.vaults(vault_).owner;
+        require(account != address(0), 'No owner for the vault');
+        bytes12[] storage userVault = vaults[account];
+        for (uint256 i = 0; i < userVault.length; i++) {
+            require(userVault[i] != vault_, 'already added');
```
alcueca, line 119: Do we need to revert? Would be better to just return?

Both options are fine, but let's be consistent here and in `removeVault`, and consult with frontend engineering.

contracts/utils/convex/ConvexYieldWrapper.sol
```diff
@@ -0,0 +1,136 @@
...
+    }
+
+    /// @notice Adds a vault to the user's vault list
+    /// @param vaultId The id of the vault being added
+    function addVault(bytes12 vaultId) external {
+        address account = cauldron.vaults(vaultId).owner;
+        require(account != address(0), 'No owner for the vault');
+        bytes12[] storage vaults_ = vaults[account];
+        for (uint256 i = 0; i < vaults_.length; i++) {
+            require(vaults_[i] != vaultId, 'already added');
```
alcueca, line 56: Do we need to revert, or should we just return and do nothing if the cvault has already been added? Check with frontend engineering and make sure the implementation is consistent with `removeVault`

alcueca (approved):

PR #330, opened 2021-12-07
Reward calling auction on the Witch

alcueca:
Simple refactor of the Witch to add the option of rewarding those that call `auction` with a portion of the collateral in the vault.


devtooligan:
closing and tracking it on an issue

PR #333, opened 2021-12-14
Disable fyToken flash loans by default

alcueca:
*Summary*
Disable flash loans by default on fyToken, following the Join pattern.

*Details*
We can be subject to DoS attacks when releasing new series, particularly if doing so at the same time that we roll a strategy onto them.

If we deploy a pool and we don't roll a strategy onto it in the same transaction, a malicious actor can skew the ratio in the pool using the free fyToken flash loans, just to make our slippage protection kick in and revert the governance action. While no funds would be lost, service continuity would be affected.

This PR sets the fyToken flash loans to be disabled by default, along with allowing to charge a fee in case that would be necessary to protect from unforeseen attacks.

### DIFF
--- a/contracts/FYToken.sol
+++ b/contracts/FYToken.sol
@@ -23,6 +23,7 @@ contract FYToken is IFYToken, IERC3156FlashLender, AccessControl(), ERC20Permit,
     using CastU256U32 for uint256;
 
     event Point(bytes32 indexed param, address value);
+    event FlashFeeFactorSet(uint256 indexed fee);
     event SeriesMatured(uint256 chiAtMaturity);
     event Redeemed(address indexed from, address indexed to, uint256 amount, uint256 redeemed);
 
@@ -30,6 +31,8 @@ contract FYToken is IFYToken, IERC3156FlashLender, AccessControl(), ERC20Permit,
 
     uint256 constant internal MAX_TIME_TO_MATURITY = 126144000; // seconds in four years
     bytes32 constant internal FLASH_LOAN_RETURN = keccak256("ERC3156FlashBorrower.onFlashLoan");
+    uint256 constant FLASH_LOANS_DISABLED = type(uint256).max;
+    uint256 public flashFeeFactor = FLASH_LOANS_DISABLED;       // Fee on flash loans, as a percentage in fixed point with 18 decimals. Flash loans disabled by default.
 
     IOracle public oracle;                                      // Oracle for the savings rate.
     IJoin public join;                                          // Source of redemption funds.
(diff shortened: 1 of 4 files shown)


contracts/FYToken.sol
```diff
@@ -23,13 +23,16 @@ contract FYToken is IFYToken, IERC3156FlashLender, AccessControl(), ERC20Permit,
...
     event Point(bytes32 indexed param, address value);
+    event FlashFeeFactorSet(uint256 indexed fee);
     event SeriesMatured(uint256 chiAtMaturity);
     event Redeemed(address indexed from, address indexed to, uint256 amount, uint256 redeemed);
 
     uint256 constant CHI_NOT_SET = type(uint256).max;
 
     uint256 constant internal MAX_TIME_TO_MATURITY = 126144000; // seconds in four years
     bytes32 constant internal FLASH_LOAN_RETURN = keccak256("ERC3156FlashBorrower.onFlashLoan");
+    uint256 constant FLASH_LOANS_DISABLED = type(uint256).max;
```
devtooligan, line 34: Is the reason that this disables the flash loan because it would overflow?  It took me a second to realize that,  maybe we can comment that?

devtooligan (approved):
lgtm - maybe add a comment about how the disable works

test/032_fyToken_flash.ts
```diff
@@ -54,54 +54,87 @@ describe('FYToken - flash', function () {
...
+        const fee = principal.mul(5).div(100)
+        await fyToken.mint(borrower.address, fee)
+        await expect(borrower.flashBorrow(fyToken.address, principal, actions.normal))
+          .to.emit(fyToken, 'Transfer')
+          .withArgs(borrower.address, '0x0000000000000000000000000000000000000000', principal.add(fee))
+
+        expect(await fyToken.balanceOf(owner)).to.equal(0)
+        expect(await borrower.flashBalance()).to.equal(principal.add(fee))
+        expect(await borrower.flashToken()).to.equal(fyToken.address)
+        expect(await borrower.flashAmount()).to.equal(principal)
```
sblOWPCKCR, line 123: check `flashFee` too

contracts/FYToken.sol
```diff
@@ -232,8 +252,9 @@ contract FYToken is IFYToken, IERC3156FlashLender, AccessControl(), ERC20Permit,
     {
         require(token == address(this), "Unsupported currency");
         _mint(address(receiver), amount);
+        uint128 fee = _flashFee(amount).u128();
         require(receiver.onFlashLoan(msg.sender, token, amount, 0, data) == FLASH_LOAN_RETURN, "Non-compliant borrower");
```
sblOWPCKCR, line 256: s/0/fee/

sblOWPCKCR (changes requested):

contracts/FYToken.sol
devtooligan: @alcueca

contracts/FYToken.sol
alcueca: Commented.

contracts/FYToken.sol
alcueca: Oh shit.

Oh well, that means that we can't enable the flash loans on the june series unless we do it without a fee.

sblOWPCKCR (approved):

PR #339, opened 2022-01-25
Feat/c4rena competition

devtooligan:
# questions / comments for @alcueca @iamsahu :
 - [x] setApprovals() - why do we set to 0 first?
 - [x] If there's a good reason for `ConvexStakingWrapper._getTotalSupply` we should disclose it (or else delete the fn)
 - [x] Is there a reason `ConvexStakingWrapper._getTotalSupply` is marked as "virtual"?  If yes, document, if no, then maybe remove virtual
 - [x] Fyi - looks like the CvxMining library is duped, once in its own file and once in ConvexStakingWrapper.sol.  I updated the one in ConvexStakingWrapper.sol but completely ignored the CvxMining.sol file
 - [x] I didn't have time to double check the decimals in the oracle, I'm assuming this was done in the code review
 
# Known issues / Intentional deviations from best practices
- We intentionally do not check validity of contract addresses using `isContract` or some other means
- We intentionally chose not to use Yul for minor gas savings in favor of readability
- We intentionally did not make use of custom errors in spite of the potential gas savings
- The length of each revert string is intentional and we do not wish to save gas by shortening them
- We intentionally did not apply "unchecked" to the incrementing of loop counter variables, preferring readability to small runtime gas savings
- Any "missing" address == 0 checks were intentionally omitted to save gas
- We have intentionally inlined as much logic as we care to and do not wish to save any additional runtime gas by inlining more
- We have abstracted as much logic as we care to and do not wish to reduce bytecode size to save deployment gas by abstracting more
-  In `Cvx3CrvOracle.sol`, `peek()` and `get()` both point to the same internal fn `_peek()`, however `get()` is transactional.  This is all done intentionally to emulate behavior of Yield oracles.

# Naming conventions:
- We intentionally use all-caps variable names instead of mixed case for state variables DAI, USDC, USDT, and IWETH9
- We intentionally use PascalCase for certain function names to be consistent with the [contracts](https://github.com/convex-eth/platform/blob/main/contracts/contracts/wrappers/ConvexStakingWrapperAbra.sol) these were based on
- We intentionally use snake_case for certain variable names to be consistent with the [contracts](https://github.com/convex-eth/platform/blob/main/contracts/contracts/wrappers/ConvexStakingWrapperAbra.sol) these were based on
- We intentionally use variable name `I` to represent integral to be consistent with the [contracts](https://github.com/convex-eth/platform/blob/main/contracts/contracts/wrappers/ConvexStakingWrapperAbra.sol) these were based on

# NatSpec
- We intentionally only include @notice and @dev comments for contracts and functions when we think it is helpful.  If it is omitted, it is intentional.

### DIFF
--- a/contracts/mocks/ConvexYieldWrapperMock.sol
+++ b/contracts/mocks/ConvexYieldWrapperMock.sol
@@ -206,11 +206,9 @@ contract ConvexYieldWrapperMock is ERC20, AccessControl {
         uint256 collateral;
         Balances memory balance;
         for (uint256 i = 0; i < userVault.length; i++) {
-            if (userVault[i] != bytes12(0)) {
-                if (cauldron.vaults(userVault[i]).owner == account_) {
-                    balance = cauldron.balances(userVault[i]);
-                    collateral = collateral + balance.ink;
-                }
+            if (cauldron.vaults(userVault[i]).owner == account_) {
+                balance = cauldron.balances(userVault[i]);
+                collateral = collateral + balance.ink;
             }
         }
 
--- a/contracts/utils/convex/ConvexYieldWrapper.sol
+++ b/contracts/utils/convex/ConvexYieldWrapper.sol
@@ -2,9 +2,9 @@
 
 pragma solidity 0.8.6;
 
-import '@yield-protocol/vault-interfaces/ICauldron.sol';
-import '@yield-protocol/vault-interfaces/DataTypes.sol';
-import './ConvexStakingWrapper.sol';
+import "@yield-protocol/vault-interfaces/ICauldron.sol";
+import "@yield-protocol/vault-interfaces/DataTypes.sol";
+import "./ConvexStakingWrapper.sol";
 
 /// @title Convex staking wrapper for Yield platform
 /// @notice Enables use of convex LP positions as collateral while still receiving rewards
@@ -50,11 +50,12 @@ contract ConvexYieldWrapper is ConvexStakingWrapper {
     /// @param vaultId The id of the vault being added
     function addVault(bytes12 vaultId) external {
         address account = cauldron.vaults(vaultId).owner;
-        require(account != address(0), 'No owner for the vault');
+        require(account != address(0), "No owner for the vault");
         bytes12[] storage vaults_ = vaults[account];
         uint256 vaultsLength = vaults_.length;
+
         for (uint256 i = 0; i < vaultsLength; i++) {
-            require(vaults_[i] != vaultId, 'already added');
+            require (vaults_[i] != vaultId, "Vault already added");
         }
         vaults_.push(vaultId);
         vaults[account] = vaults_;
@@ -69,6 +70,7 @@ contract ConvexYieldWrapper is ConvexStakingWrapper {
         if (account != owner) {
             bytes12[] storage vaults_ = vaults[account];
             uint256 vaultsLength = vaults_.length;
+            bool found;
             for (uint256 i = 0; i < vaultsLength; i++) {
                 if (vaults_[i] == vaultId) {
                     bool isLast = i == vaultsLength - 1;
@@ -76,10 +78,12 @@ contract ConvexYieldWrapper is ConvexStakingWrapper {
                         vaults_[i] = vaults_[vaultsLength - 1];
                     }
                     vaults_.pop();
+                    found = true;
                     emit VaultRemoved(account, vaultId);
                     break;
                 }
             }
+            require (found, "Vault not found");
             vaults[account] = vaults_;
         }
     }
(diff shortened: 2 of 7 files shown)


contracts/utils/convex/ConvexYieldWrapper.sol
```diff
@@ -67,9 +68,14 @@ contract ConvexYieldWrapper is ConvexStakingWrapper {
         address owner = cauldron.vaults(vaultId).owner;
         if (account != owner) {
             bytes12[] storage vaults_ = vaults[account];
-            for (uint256 i = 0; i < vaults_.length; i++) {
+            uint256 vaultsLength = vaults_.length;
+            for (uint256 i = 0; i < vaultsLength; i++) {
                 if (vaults_[i] == vaultId) {
-                    vaults_[i] = bytes12(0);
+                    bool isLast = i == vaultsLength - 1;
```
devtooligan, line 74: haha - this contradicts the readme

contracts/utils/convex/ConvexYieldWrapper.sol
```diff
@@ -91,12 +97,11 @@ contract ConvexYieldWrapper is ConvexStakingWrapper {
         //add up all balances of all vaults
         uint256 collateral;
         DataTypes.Balances memory balance;
-        for (uint256 i = 0; i < userVault.length; i++) {
-            if (userVault[i] != bytes12(0)) {
```
devtooligan, line 95: if we revert the array cleanup in removeVault then we should add this back

contracts/mocks/ConvexYieldWrapperMock.sol
```diff
@@ -206,11 +206,9 @@ contract ConvexYieldWrapperMock is ERC20, AccessControl {
         uint256 collateral;
         Balances memory balance;
         for (uint256 i = 0; i < userVault.length; i++) {
-            if (userVault[i] != bytes12(0)) {
```
devtooligan, line 209: if we revert the array cleanup in removeVault then we should add this back

alcueca:
On `setApprovals` I suspect that they were reading about the approval race condition and decided to go nuts. That attack is stupid to start with, and only allows the attacker to get a higher approval than he is granted. Here the attacker would need to be `convexBooster` or `convexPool`, and we are granting them max approval anyway.

Axe this one.

alcueca:
The only point that I see to `_getTotalSupply()` is so that it can be overriden by inheriting functions, without having to change the code in `earned`, `checkpoint` and `_checkpointAndClaim`. I would let that one be.

alcueca:
On CvxMining, let's keep it as a separate file.

contracts/utils/convex/ConvexYieldWrapper.sol
alcueca: I like the cleanup

contracts/mocks/ConvexYieldWrapperMock.sol
alcueca: I like the cleanup

alcueca:
@Eihcir0, please have a look at my commits to make sure I didn't mess up.

Also, I thought that `>` was cheaper in gas than `!=`, do I get it wrong?

alcueca (approved):
LGTM, someone please review my commits.

devtooligan:
> Also, I thought that `>` was cheaper in gas than `!=`, do I get it wrong?

In my testing there is really not much of a difference at all, and in fact, in the case of x != 0, from what I understand it actually gets compiled to x > 0  -- in other cases the difference was a whopping 3 gas.  Howeve, during my review of the c4 findings this was one of the top reported “gas optimizations”.  I guess we can leave it in for now to save everybody some time.  Personally I don’t have a preference either way, we could also leave a note about it in the readme and leave it as x > 0

The reason I changed it yesterday was because it had been a while since I looked into this issues  and I just forgot there was no diffrerence in gas.  I had just finished reading like 20 “findings” that suggested they should use x != 0  so it probably brainwashed me.   If I would’ve remembered that I wouldnt have changed it.  We can change it back now, but if we do I would add a line to the readme saying we are intentionally doing so

Alberto:
> Nah, leave it as is and put a note in the README

👍

contracts/utils/convex/ConvexYieldWrapper.sol
```diff
@@ -133,4 +148,43 @@ contract ConvexYieldWrapper is ConvexStakingWrapper {
...
+            IRewardStaking(convexPool).withdraw(balance_, true);
+
+            // Transfer the withdrawn convex tokens to rescue address
+            IERC20(convexToken).safeTransfer(rescueAddress_, balance_);
+
+            // Transfer the reward tokens if any to the rescueAddress
+            uint256 rewardCount = rewards.length;
+            for (uint256 i = 0; i < rewardCount; i++) {
+                RewardType storage reward = rewards[i];
+                uint256 rewardToken = reward.reward_token;
```
iamsahu, line 182: This needs to be address & not uint256

Issue #341, opened 2022-02-04
Malicious Users Can Duplicate Protocol Earned Yield By Transferring wCVX Tokens To Another Account


iamsahu:
## Report

- https://github.com/code-423n4/2022-01-yield-findings/issues/86
- https://github.com/code-423n4/2022-01-yield-findings/issues/85

## Proposed fix
Reinstate `_beforeTokenTransfer`, or override `_transfer` to use `checkpoint` except when transferring between wrapper and join

alcueca:
Has this been fixed, @iamsahu ?

iamsahu:
This was fixed by completely removing the Wrapper token.

PR #349, opened 2022-02-06
[c4-2022-01-86] Fix for issue 341

iamsahu:
## Goal
Fix the #341 
## Details
`_transfer` was overriden to add conditional `_checkpoint`. The conditions under which a checkpoint happens:

1. The transfer is not between `join` & the `wrapper`
2. When the destination of the transfer is the `wrapper` the checkpoint is done only for the non-wrapper address. 
 
We have implemented 2 to prevent the wrapper from getting any rewards. Here is an example showing how the `wrapper` contract ends up receiving a reward when the above check is not present.
1. Wrapped token is transferred to the `wrapper` by `User A`
2. `User B` makes another wrapped token transfer to the `wrapper`. This transaction would lead to the `wrapper` receiving rewards in proportion to the balance that was transferred by `User A`. This happens because while checkpointing for the `wrapper` there would be balance available for the wrapper leading it to receiving som reward


alcueca:
There is another issue that would make `_checkpoint()` take only one parameter.

Maybe then it would be easier to do:
```
if src != wrapper && src != join: _checkpoint(src)
if dst != wrapper && dst != join: _checkpoint(dst)
```

alcueca (approved):
LGTM

Issue #351, opened 2022-02-06
ConvexYieldWrapper: Add vault doesn't have a collateral check


iamsahu:
## Problem
Right now a vault belonging to a user could be added whether it is for the same convex collateral or something else
## Proposed fix
Add a check to verify if the vault being added is for the same collateral

Sabnock01:
Should this issue not be closed as it appears to be fixed by PR #352 ?

PR #352, opened 2022-02-06
Prevent vault of different collateral to be added

iamsahu:
Fix for #351 

## Details
A check was added in the `addVault` function to prevent a vault for a different ilk to be added

## Tests
A test was added in the `092_convex_wrapper` to check if the vault addition fails if a different vault is added.


alcueca:
Something seems off with the single quotes and the TODOs, would you check you branched off the latest code?

test/092_convex_wrapper.ts
```diff
@@ -304,4 +304,19 @@ describe('Convex Wrapper', async function () {
...
+    var join = await ladle.joins(CVX3CRV)
+    await convexWrapper.point(join)
+    // Batch action to build a vault & add it to the wrapper
+    const addVaultCall = convexLadleModule.interface.encodeFunctionData('addVault', [
+      convexWrapper.address,
+      '0x000000000000000000000000',
+    ])
+    await expect(
+      ladle.batch([ladle.buildAction(seriesId, USDC), ladle.moduleCallAction(convexLadleModule.address, addVaultCall)])
+    ).to.be.revertedWith('revert Vault is for different ilk')
```
alcueca, line 320: The error message is different:
"Vault is for different ilk"
'revert Vault is for different ilk'

Is this test ok?

alcueca (commented):
Looks good, but I'm concerned on the test giving a false positive. Would you please check?

test/092_convex_wrapper.ts
iamsahu: Really strange that it is passing. Let me investigate.

iamsahu: The above test is failing locally. After changing the message to reflect what is in the contract is making it pass.

alcueca (approved):
It's concerning that our CI doesn't catch all test errors, though :/

Maybe moving to foundry would fix this.

Issue #354, opened 2022-02-06
ConvexYieldWrapper: wrap function could be exploited `wrap` tokens & send it to a desired address


iamsahu:
## Goal
Remove exploitation possible on `wrap`.
## Proposed fix
Remove the `to` address from the `wrap` function and send it always to the `join`

PR #355, opened 2022-02-06
Removed to from `wrap` in ConvexYieldWrapper

iamsahu:
Fix for issue #354 

## Details
Modified the wrap function to always send the wrapped token to the `join`


alcueca (approved):
:+1:

PR #356, opened 2022-02-07
Fix multiple small issues

iamsahu:
Fixed the following issues from the code4rena findings.
- https://github.com/code-423n4/2022-01-yield-findings/issues/8
- https://github.com/code-423n4/2022-01-yield-findings/issues/33
- https://github.com/code-423n4/2022-01-yield-findings/issues/43
- https://github.com/code-423n4/2022-01-yield-findings/issues/49
- https://github.com/code-423n4/2022-01-yield-findings/issues/54
- https://github.com/code-423n4/2022-01-yield-findings/issues/62
- https://github.com/code-423n4/2022-01-yield-findings/issues/113
- https://github.com/code-423n4/2022-01-yield-findings/issues/63


contracts/utils/convex/ConvexYieldWrapper.sol
```diff
@@ -159,6 +158,7 @@ contract ConvexYieldWrapper is ConvexStakingWrapper {
         address destination_
     ) external auth {
         require(amount_ != 0, "amount is 0");
+        require(token_ != convexToken, "cannot withdraw wrappable token");
```
alcueca, line 161: I'm not so sure anymore about this one. If the wrapper is not expected to hold any tokens, except received rewards, then we wouldn't need this. Let's discuss.

alcueca (commented):
On thing to discuss.

contracts/utils/convex/ConvexYieldWrapper.sol
iamsahu: You are right. Changing this.

iamsahu: Do you think we should add a protection to prevent the protocol from transferring the wrapped token.
For eg: The liquidator would be transferring the token to the contract & then call unwrap to get the asset. It might be the case that this gets executed in 2 single transaction which theoretically gives protocol a window of opportunity to exploit and claim the token.

alcueca: No liquidator would do that. I think we can just assume that wrapper tokens should never be left in the contract, as we do with the rest of the protocol.

alcueca (approved):
Thanks

PR #357, opened 2022-02-07
Fix to optimize `addVault` & `removeVault` code in ConvexYieldWrapper

iamsahu:
Fixes for:
- https://github.com/code-423n4/2022-01-yield-findings/issues/111
- https://github.com/code-423n4/2022-01-yield-findings/issues/107


alcueca (approved):
Looks good to me :+1:

PR #359, opened 2022-02-07
Optimized `_peek` to save gas in `Cvx3CrvOracle`

iamsahu:
Fix the issue [70](https://github.com/code-423n4/2022-01-yield-findings/issues/70)
Also, renamed test file from `018_oracles_convex` to `020_oracles_convex`.


alcueca (approved):
LGTM :+1:

PR #361, opened 2022-02-08
Fix: composite oracle updateTime bug

### LINKED ISSUE #360: fix: compositeOracle returns 0 for updateTime
Due to an error in the logic for `peek()`, `get()` and `_peek()` this oracle always returns 0 for updateTime.

The proposed fix is to initialize `updateTime` to type(uint256).max and then revert if updateTime is not <= block.timestamp

devtooligan:
closes #360


contracts/oracles/composite/CompositeMultiOracle.sol
```diff
@@ -31,74 +35,119 @@ contract CompositeMultiOracle is IOracle, AccessControl {
...
-    {
-        bytes6[] memory reverse = new bytes6[](path.length);
+    /// @notice Set or reset an price path and its reverse path
+    /// @param base Id of base token
+    /// @param quote Id of quote token
+    /// @param path Path from base to quote
+    function setPath(
+        bytes6 base,
+        bytes6 quote,
+        bytes6[] memory path
```
alcueca, line 45: Should this be `calldata`? Or not because we are assigning it to `paths[base][quote]` below?

contracts/oracles/composite/CompositeMultiOracle.sol
```diff
@@ -31,74 +35,119 @@ contract CompositeMultiOracle is IOracle, AccessControl {
...
+    ) external auth {
+        uint256 pathLength = path.length;
+        bytes6[] memory reverse = new bytes6[](pathLength);
         bytes6 base_ = base;
-        for (uint256 p = 0; p < path.length; p++) {
-            require (sources[base_][path[p]] != IOracle(address(0)), "Source not found");
-            base_ = path[p];
-            reverse[path.length - (p + 1)] = base_;
+        unchecked {
+            for (uint256 p = 0; p < pathLength; ++p) {
```
alcueca, line 51:
```suggestion
            for (uint256 p; p < pathLength; ++p) {
```

contracts/oracles/composite/CompositeMultiOracle.sol
```diff
@@ -31,74 +35,119 @@ contract CompositeMultiOracle is IOracle, AccessControl {
...
+    /// @notice Convert amountBase base into quote at the latest oracle price, through a path is exists.
+    /// @param base Id of base token
+    /// @param quote Id of quote token
+    /// @param amountBase Amount of base to convert to quote
+    /// @return amountQuote Amount of quote token converted from base
+    function peek(
+        bytes32 base,
+        bytes32 quote,
+        uint256 amountBase
+    ) external view virtual override returns (uint256 amountQuote, uint256 updateTime) {
```
alcueca, line 72: I like to break the `returns` into its own line, but that's just me.

contracts/oracles/composite/CompositeMultiOracle.sol
```diff
@@ -31,74 +35,119 @@ contract CompositeMultiOracle is IOracle, AccessControl {
...
         amountQuote = amountBase;
         bytes6 base_ = base.b6();
         bytes6 quote_ = quote.b6();
         bytes6[] memory path = paths[base_][quote_];
-        for (uint256 p = 0; p < path.length; p++) {
-            (amountQuote, updateTime) = _peek(base_, path[p], amountQuote, updateTime);
-            base_ = path[p];
+        uint256 pathLength = path.length;
+        unchecked {
+            for (uint256 p = 0; p < pathLength; ++p) {
```
alcueca, line 80:
```suggestion
            for (uint256 p; p < pathLength; ++p) {
```

contracts/oracles/composite/CompositeMultiOracle.sol
```diff
@@ -31,74 +35,119 @@ contract CompositeMultiOracle is IOracle, AccessControl {
...
         amountQuote = amountBase;
         bytes6 base_ = base.b6();
         bytes6 quote_ = quote.b6();
         bytes6[] memory path = paths[base_][quote_];
-        for (uint256 p = 0; p < path.length; p++) {
-            (amountQuote, updateTime) = _get(base_, path[p], amountQuote, updateTime);
-            base_ = path[p];
+        uint256 pathLength = path.length;
+        unchecked {
+            for (uint256 p = 0; p < pathLength; ++p) {
```
alcueca, line 107:
```suggestion
            for (uint256 p; p < pathLength; ++p) {
```

test/013_oracles_composite.ts
```diff
@@ -97,6 +102,15 @@ describe('Oracles - Composite', function () {
...
 
+  it('reverts on timestamp greater than current block', async () => {
+    await daiEthAggregator.setTimestamp(
+      BigNumber.from('0xffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffff')
+    ) // 1 DAI (1^18) in ETH
+    await expect(compositeMultiOracle.peek(bytes6ToBytes32(DAI), bytes6ToBytes32(ETH), WAD)).to.be.revertedWith(
+      'Invalid updateTime'
+    )
+  })
+
```
alcueca, line 113: Missing test for picking the oldest updateTime in a path.

test/013_oracles_composite.ts
```diff
@@ -82,10 +83,14 @@ describe('Oracles - Composite', function () {
     await compositeMultiOracle.setPath(DAI, USDC, [ETH])
```
alcueca, line 83: `setPath` and `setSource` are actually untested. Please do so, including testing that only authorized users can execute them.

alcueca (changes requested):
Thanks for the fix, and for going the extra mile and adding proper natspec.

I added a small gas golfing thing, and requested a few more tests to get proper coverage.

contracts/oracles/composite/CompositeMultiOracle.sol
devtooligan: Good catch.  I didnt see where we assigning it below, and no tests broke when i switched it so 👍

contracts/oracles/composite/CompositeMultiOracle.sol
devtooligan: I agree it looks nicer but Prettier keeps moving it back up after I change it so until i figure out how to change the prettier rule, I will leave it

test/013_oracles_composite.ts
devtooligan: 👍

alcueca (approved):
Thank you good ser :pray:

PR #363, opened 2022-02-08
Removed unnecessary += to save gas

iamsahu:
Fix for issue: https://github.com/code-423n4/2022-01-yield-findings/issues/101


alcueca (approved):
Wohooo

PR #365, opened 2022-02-09
Feat/347 rewrite witch

devtooligan:


devtooligan:
@alcueca One failing test, actually one failing assertion, off by 1.  All the other tests are passing.

Screen Shot 2022-02-11 at 11 34 24 AM

contracts/Witch.sol
```diff
@@ -69,138 +75,205 @@ contract Witch is AccessControl() {
...
-        auctions[vaultId] = Auction({
-            owner: vault_.owner,
-            start: block.timestamp.u32()
-        });
+        auctions[vaultId] = Auction({owner: vault_.owner, start: block.timestamp.u32()});
         cauldron.give(vaultId, address(this));
         emit Auctioned(vaultId, block.timestamp.u32());
     }
 
+    /// @dev Private fn called by buy() and payAll()
```
alcueca, line 121: What does the function do, though?

contracts/Witch.sol
```diff
@@ -69,138 +75,205 @@ contract Witch is AccessControl() {
...
+            elapsed = uint32(block.timestamp) - auctionStart;
+        }
+
+        uint256 t = elapsed > duration ? 1e18 : elapsed.wdivup(duration);
+
+        //          (      a         )
+        // inkOut = (artIn / totalArt) * totalInk * (p + (1 - p) * t)
+        {
+            uint256 a = uint256(artIn).wdivup(totalArt);
+            inkOut = a.wmul(totalInk).wmulup(uint256(p) + uint256(1e18 - p).wmulup(t));
```
alcueca, line 158: Do you need to round up in any arithmetic operation in this function?

contracts/Witch.sol
```diff
@@ -69,138 +75,205 @@ contract Witch is AccessControl() {
...
-    function buy(bytes12 vaultId, uint128 base, uint128 min)
-        external
-        returns (uint256 ink)
-    {
-        require (auctions[vaultId].start > 0, "Vault not under auction");
+    /// @param vaultId Id of vault to buy
+    /// @param base Amount of base to pay
+    /// @param min Minimum amount of collateral that must be received
+    /// @return ink Amount of vault collateral sold
+    function buy(
```
alcueca, line 172: You can do an internal `_buy` function that takes the vault and balances structs as parameters, along with whatever else is needed. Then `paySome` and `payAll` would be thin functions that just read a couple of structs and call the single `_buy` function that has the logic.

alcueca (changes requested):
Let's think a bit about rounding, and about sharing common functionality between functions.

contracts/Witch.sol
devtooligan: [Added natspec comment](https://github.com/yieldprotocol/vault-v2/pull/365/commits/a9913aa9dc6e9e9a2f48c9753844ad94156a0a0e), is that good enough?  Or would you like me to refactor or perhaps inline this logic in the main fn's inside of a {} block?

contracts/Witch.sol
devtooligan: I did my best to stick with the rounding from the original function.  Pretty much everything was a straight port accept for the initial term.  Re: the initial term, the old version was:
```
totalnk.divup(totalArk).mulup(<next term>...
```

so I interpret the new fn as:
```
(artIn.divup(totalArt)).mulup(<next term>...
```

WDYT?

contracts/Witch.sol
devtooligan: hmm actually this fn may go when I implement your `_buy()` suggestion below.

contracts/Witch.sol
```diff
@@ -217,55 +217,10 @@ contract Witch is AccessControl {
...
- paid is settled up with the join.             `.     ``-----'-''     ,'
-                                                  -.               ,-
-                                                     `-._______.-'gpyy
-
-
-*/
-
-    /// @notice Remove debt from a vault, and return how much collateral should be given out.
-    /// Auction limits apply.
-    /// @dev If the debt is returned to zero, the vault is returned to its original owner.
```
devtooligan: @alcueca u got rid of the ascii art?  i thot it was helpful!

contracts/Witch.sol
```diff
@@ -217,55 +217,10 @@ contract Witch is AccessControl {
...
- and the collateral from the vault!      \\ ;'      \\                   `: //
-                                          `(          \\                   )'
- The Ladle then dishes out the collateral   :.          \\,----,         ,;
- to you, dear user.  And the base you        `.`--.___   (    /  ___.--','
- paid is settled up with the join.             `.     ``-----'-''     ,'
-                                                  -.               ,-
-                                                     `-._______.-'gpyy
-
-
-*/
```
devtooligan: @alcueca u got rid of the ascii art?  i thot it was helpful!

alcueca:
Continues in #411

PR #366, opened 2022-02-09
ConvexYieldWrapper: modified for loop initialization to save gas

iamsahu:
Fix for:
- https://github.com/code-423n4/2022-01-yield-findings/issues/14


alcueca (approved):
:golfing:

Issue #343, opened 2022-02-04
Rewards distribution can be disrupted by a early user


iamsahu:
## Report
https://github.com/code-423n4/2022-01-yield-findings/issues/116
## Proposed Fix
Fix use of casting

PR #367, opened 2022-02-09
ConvexYieldWrapper: added unchecked blocks for gas saving

iamsahu:
Fixes for the following:

- https://github.com/code-423n4/2022-01-yield-findings/issues/67
- https://github.com/code-423n4/2022-01-yield-findings/issues/65
- https://github.com/code-423n4/2022-01-yield-findings/issues/64
- https://github.com/code-423n4/2022-01-yield-findings/issues/112
- #343


alcueca:
Let's take a different approach to this one.

Only `reward_integral` and `reward_remaining` are uint128.

Upcasts from uint128 to uint256 are automatic. Meaning that `uint128(x) + uint256(y)` gets casted into `uint256(x) + uint256(y)` by the compiler. We can remove upcasts from the code to make it clearer when there is a mix of uint256 and uint128.

Downcasts to uint128 should happen *only* at the last possible operation, when assigning the value to storage. At that point we will use one of our SafeCast operations (`.u128()`)

alcueca (approved):
Looks good to me, much cleaner.

PR #368, opened 2022-02-09
ConvexYieldWrapper: modified earned function to the latest one

iamsahu:
Fix for the issue:
- https://github.com/code-423n4/2022-01-yield-findings/issues/95

### DIFF
--- a/contracts/utils/convex/ConvexStakingWrapper.sol
+++ b/contracts/utils/convex/ConvexStakingWrapper.sol
@@ -103,21 +103,18 @@ contract ConvexStakingWrapper is ERC20 {
             RewardType storage reward = rewards.push();
             reward.reward_token = crv;
             reward.reward_pool = mainPool;
-            reward.reward_integral = 0;
-            reward.reward_remaining = 0;
 
             reward = rewards.push();
             reward.reward_token = cvx;
-            reward.reward_pool = address(0); // This is set to address(0) as initially we don't know if the pool has cvx rewards
-            reward.reward_integral = 0;
-            reward.reward_remaining = 0;
+            // The reward_pool is set to address(0) as initially we don't know if the pool has cvx rewards.
+            // And since the default is address(0) we don't explicitly set it
 
             registeredRewards[crv] = CRV_INDEX + 1; //mark registered at index+1
             registeredRewards[cvx] = CVX_INDEX + 1; //mark registered at index+1
         }
 
         uint256 extraCount = IRewardStaking(mainPool).extraRewardsLength();
-        for (uint256 i = 0; i < extraCount; i++) {
+        for (uint256 i; i < extraCount; ++i) {
             address extraPool = IRewardStaking(mainPool).extraRewards(i);
             address extraToken = IRewardStaking(extraPool).rewardToken();
             if (extraToken == cvx) {
@@ -128,8 +125,6 @@ contract ConvexStakingWrapper is ERC20 {
                 RewardType storage reward = rewards.push();
                 reward.reward_token = extraToken;
                 reward.reward_pool = extraPool;
-                reward.reward_integral = 0;
-                reward.reward_remaining = 0;
 
                 registeredRewards[extraToken] = rewards.length; //mark registered at index+1
             }
@@ -275,7 +270,7 @@ contract ConvexStakingWrapper is ERC20 {
         uint256 rewardCount = rewards.length;
         claimable = new EarnedData[](rewardCount);
 
-        for (uint256 i = 0; i < rewardCount; i++) {
+        for (uint256 i; i < rewardCount; ++i) {
             RewardType storage reward = rewards[i];
 
             if (reward.reward_pool == address(0)) {
--- a/contracts/utils/convex/ConvexStakingWrapper.sol
+++ b/contracts/utils/convex/ConvexStakingWrapper.sol
@@ -27,11 +27,6 @@ contract ConvexStakingWrapper is ERC20 {
         mapping(address => uint256) claimable_reward;
     }
 
-    uint256 public cvx_reward_integral;
-    uint256 public cvx_reward_remaining;
-    mapping(address => uint256) public cvx_reward_integral_for;
-    mapping(address => uint256) public cvx_claimable_reward;
-
     //constants/immutables
     uint256 public convexPoolId;
     address public constant convexBooster = address(0xF403C135812408BFbE8713b5A23a04b3D48AAE31);
@@ -54,6 +49,7 @@ contract ConvexStakingWrapper is ERC20 {
 
     //rewards
     RewardType[] public rewards;
+    mapping(address => uint256) public registeredRewards;
 
     event Deposited(address indexed _user, address indexed _account, uint256 _amount, bool _wrapped);
     event Withdrawn(address indexed _user, uint256 _amount, bool _unwrapped);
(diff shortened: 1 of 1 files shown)


contracts/utils/convex/ConvexStakingWrapper.sol
```diff
@@ -100,23 +99,40 @@ contract ConvexStakingWrapper is ERC20 {
...
-        uint256 rewardsLength = rewards.length;
-
-        if (rewardsLength == 0) {
+        if (rewards.length == 0) {
             RewardType storage reward = rewards.push();
             reward.reward_token = crv;
             reward.reward_pool = mainPool;
-            rewardsLength = 1;
+            reward.reward_integral = 0;
+            reward.reward_remaining = 0;
```
alcueca, line 107: The two lines above can be removed.

contracts/utils/convex/ConvexStakingWrapper.sol
```diff
@@ -100,23 +99,40 @@ contract ConvexStakingWrapper is ERC20 {
...
+
+            reward = rewards.push();
+            reward.reward_token = cvx;
+            reward.reward_pool = address(0); // This is set to address(0) as initially we don't know if the pool has cvx rewards
```
alcueca, line 111: This line can be replaced by just the comment.

contracts/utils/convex/ConvexStakingWrapper.sol
```diff
@@ -100,23 +99,40 @@ contract ConvexStakingWrapper is ERC20 {
...
+            reward.reward_integral = 0;
+            reward.reward_remaining = 0;
```
alcueca, line 113: The two lines above can be removed.

contracts/utils/convex/ConvexStakingWrapper.sol
```diff
@@ -100,23 +99,40 @@ contract ConvexStakingWrapper is ERC20 {
...
+
+            registeredRewards[crv] = CRV_INDEX + 1; //mark registered at index+1
+            registeredRewards[cvx] = CVX_INDEX + 1; //mark registered at index+1
```
alcueca, line 116: Why +1? Can't you register them at the index, and then subtract 1 to rewards.length when adding marks for other tokens?

contracts/utils/convex/ConvexStakingWrapper.sol
```diff
@@ -100,23 +99,40 @@ contract ConvexStakingWrapper is ERC20 {
...
         }
 
         uint256 extraCount = IRewardStaking(mainPool).extraRewardsLength();
-
-        uint256 startIndex = rewardsLength - 1;
-        for (uint256 i = startIndex; i < extraCount; i++) {
+        for (uint256 i = 0; i < extraCount; i++) {
```
alcueca, line 120: Can we remove the use of `startIndex`?

contracts/utils/convex/ConvexStakingWrapper.sol
```diff
@@ -100,23 +99,40 @@ contract ConvexStakingWrapper is ERC20 {
...
             address extraPool = IRewardStaking(mainPool).extraRewards(i);
-            RewardType storage reward = rewards.push();
-            reward.reward_token = IRewardStaking(extraPool).rewardToken();
-            reward.reward_pool = extraPool;
+            address extraToken = IRewardStaking(extraPool).rewardToken();
+            if (extraToken == cvx) {
+                //update cvx reward pool address
+                rewards[CVX_INDEX].reward_pool = extraPool;
+            } else if (registeredRewards[extraToken] == 0) {
```
alcueca, line 126: This means that the CVX pool can be updated, but not the pool for other tokens. That is inconsistent.

contracts/utils/convex/ConvexStakingWrapper.sol
```diff
@@ -100,23 +99,40 @@ contract ConvexStakingWrapper is ERC20 {
...
+                //add new token to list
+                RewardType storage reward = rewards.push();
+                reward.reward_token = extraToken;
+                reward.reward_pool = extraPool;
+                reward.reward_integral = 0;
+                reward.reward_remaining = 0;
```
alcueca, line 132: The two lines above can be removed.

contracts/utils/convex/ConvexStakingWrapper.sol
```diff
@@ -51,6 +49,7 @@ contract ConvexStakingWrapper is ERC20 {
 
     //rewards
     RewardType[] public rewards;
+    mapping(address => uint256) public registeredRewards;
```
alcueca, line 52: What's this for?

contracts/utils/convex/ConvexStakingWrapper.sol
```diff
@@ -308,15 +271,23 @@ contract ConvexStakingWrapper is ERC20 {
     /// @return claimable Array of earned tokens and their amount
     function earned(address _account) external view returns (EarnedData[] memory claimable) {
         uint256 supply = _getTotalSupply();
+        // uint256 depositedBalance = _getDepositedBalance(_account);
         uint256 rewardCount = rewards.length;
-        claimable = new EarnedData[](rewardCount + 1);
+        claimable = new EarnedData[](rewardCount);
```
alcueca, line 276: Mmm, why?

contracts/utils/convex/ConvexStakingWrapper.sol
```diff
@@ -308,15 +271,23 @@ contract ConvexStakingWrapper is ERC20 {
...
 
-        for (uint256 i; i < rewardCount; ++i) {
+        for (uint256 i = 0; i < rewardCount; i++) {
```
alcueca, line 278: No need to initialize `i` if it is zero.

In my implementation I initialized it to `1`, and did the `cvx` block before the loop. Now I see that I might be missing something.

contracts/utils/convex/ConvexStakingWrapper.sol
```diff
@@ -308,15 +271,23 @@ contract ConvexStakingWrapper is ERC20 {
...
             RewardType storage reward = rewards[i];
-            address rewardToken = reward.reward_token;
+
+            if (reward.reward_pool == address(0)) {
+                //cvx reward may not have a reward pool yet
+                //so just add whats already been checkpointed
```
alcueca, line 283: Don't we add what's already been checkpointed if cvx has a reward pool?

alcueca (changes requested):
Very interesting refactor. I've got lots of questions before I understand all of it, though.

contracts/utils/convex/ConvexStakingWrapper.sol
iamsahu: If we do this then if there is any new claimable reward that is there for `cvx` would be missed. The one calculated [here](https://github.com/yieldprotocol/vault-v2/blob/45256b8aca37b23b306b2a09666a5e324e14e0a9/contracts/utils/convex/ConvexStakingWrapper.sol#L299)

contracts/utils/convex/ConvexStakingWrapper.sol
iamsahu: Used to keep track of the rewards that are added. Makes the `addRewards` robust & prevent any duplication.

contracts/utils/convex/ConvexStakingWrapper.sol
```diff
@@ -100,23 +99,35 @@ contract ConvexStakingWrapper is ERC20 {
...
-        for (uint256 i = startIndex; i < extraCount; i++) {
+        for (uint256 i; i < extraCount; ++i) {
             address extraPool = IRewardStaking(mainPool).extraRewards(i);
-            RewardType storage reward = rewards.push();
-            reward.reward_token = IRewardStaking(extraPool).rewardToken();
-            reward.reward_pool = extraPool;
+            address extraToken = IRewardStaking(extraPool).rewardToken();
+            if (extraToken == cvx) {
+                //update cvx reward pool address
+                rewards[CVX_INDEX].reward_pool = extraPool;
```
alcueca, line 122:
```suggestion
                if (rewards[CVX_INDEX].reward_pool == address(0)) {
                    rewards[CVX_INDEX].reward_pool = extraPool;
                }
```

alcueca (changes requested):
Let's ship 🚢

alcueca (approved):
Oops, wrong button

PR #369, opened 2022-02-09
ConvexYieldWrapper: throw error when zero address passed in removeVault

iamsahu:
Fix for:

- https://github.com/code-423n4/2022-01-yield-findings/issues/45


alcueca (changes requested):
Actually, the finding is wrong. If someone calls `removeVault(0, 0)` we will do `owner = address(0)` in L74. Then in L75 (of the original code) `require(account != owner)` will revert.

The new L75 is not needed, however, the require message is reversed, it should be:
`require(account != owner, "vault belongs to account");`

alcueca (approved):
:+1:

PR #370, opened 2022-02-09
ConvexYieldWrapper: removed unused named returns

iamsahu:
Fix for:
- https://github.com/code-423n4/2022-01-yield-findings/issues/60

### DIFF
--- a/contracts/oracles/convex/Cvx3CrvOracle.sol
+++ b/contracts/oracles/convex/Cvx3CrvOracle.sol
@@ -79,7 +79,7 @@ contract Cvx3CrvOracle is IOracle, AccessControl {
         bytes32 base,
         bytes32 quote,
         uint256 baseAmount
-    ) external view virtual override returns (uint256 quoteAmount, uint256 updateTime) {
+    ) external view virtual override returns (uint256, uint256) {
         return _peek(base.b6(), quote.b6(), baseAmount);
     }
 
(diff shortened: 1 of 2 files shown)


contracts/oracles/convex/Cvx3CrvOracle.sol
```diff
@@ -79,7 +79,7 @@ contract Cvx3CrvOracle is IOracle, AccessControl {
         bytes32 base,
         bytes32 quote,
         uint256 baseAmount
-    ) external view virtual override returns (uint256 quoteAmount, uint256 updateTime) {
+    ) external view virtual override returns (uint256, uint256) {
         return _peek(base.b6(), quote.b6(), baseAmount);
```
alcueca, line 83: The following allows you to keep the named returns, which are nice for code parsers such as Etherscan's.
```
) external view virtual override returns (uint256 quoteAmount, uint256 updateTime) {
    (quoteAmount, updateTime) = _peek(base.b6(), quote.b6(), baseAmount);
```

contracts/oracles/convex/Cvx3CrvOracle.sol
```diff
@@ -96,7 +96,7 @@ contract Cvx3CrvOracle is IOracle, AccessControl {
         bytes32 base,
         bytes32 quote,
         uint256 baseAmount
-    ) external virtual override returns (uint256 quoteAmount, uint256 updateTime) {
+    ) external virtual override returns (uint256, uint256) {
```
alcueca, line 99: Same as above

alcueca (changes requested):
We can keep the named returns, just by assigning to them.

devtooligan (approved):
code lgtm.  confirmed we save 3 gas per redundant return ommited

Screen Shot 2022-02-14 at 8 23 13 AM

alcueca (approved):
:+1:

PR #371, opened 2022-02-10
ConvexYieldWrapper: removed the AccessControl

iamsahu:
Fix for the issue:
- https://github.com/code-423n4/2022-01-yield-findings/issues/75


alcueca (approved):
:+1:

PR #373, opened 2022-02-10
ConvexYieldWrapper: call safeTransfer directly from library

iamsahu:
Fix for the issue:
- https://github.com/code-423n4/2022-01-yield-findings/issues/72


alcueca:
Have we confirmed the gas savings of this?

PR #374, opened 2022-02-10
ConvexYieldWrapper: made removeVault pattern similar to addVault in `ConvexModule`

iamsahu:
Fix for the issue:
- https://github.com/code-423n4/2022-01-yield-findings/issues/39


alcueca (approved):
Tests would be nice

iamsahu:
I have added the tests for remove vault.

PR #375, opened 2022-02-10
ConvexYieldWrapper: Mock contracts for tests

iamsahu:
## Goal
Mock the `crv` & `convexbooster` contract such that we can use the `ConvexYieldWrapper` directly in tests rather than creating a separate `ConvexYieldWrapperMock`

### DIFF
--- a/contracts/mocks/TokenProxy.sol
+++ b/contracts/mocks/TokenProxy.sol
@@ -1,19 +1,19 @@
-// SPDX-License-Identifier: MIT
-// Original contract: https://github.com/convex-eth/platform/blob/main/contracts/contracts/wrappers/ConvexStakingWrapper.sol
+// SPDX-License-Identifier: BUSL-1.1
 pragma solidity 0.8.6;
 import "@yield-protocol/utils-v2/contracts/token/ERC20.sol";
 
+/// @notice A simple contract which could be used as a proxy for ERC20 contracts
 contract TokenProxy {
-    ERC20 immutable _curve;
+    ERC20 immutable _token;
 
-    constructor(ERC20 curve_) {
-        _curve = curve_;
+    constructor(ERC20 token_) {
+        _token = token_;
     }
 
     /// @param _owner The address from which the balance will be retrieved
     /// @return balance the balance
     function balanceOf(address _owner) external view returns (uint256 balance) {
-        return _curve.balanceOf(_owner);
+        return _token.balanceOf(_owner);
     }
 
     /// @notice send `_value` token to `_to` from `msg.sender`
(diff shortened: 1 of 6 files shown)


contracts/mocks/TokenProxy.sol
```diff
@@ -0,0 +1,54 @@
...
+    function balanceOf(address _owner) external view returns (uint256 balance) {
+        return _curve.balanceOf(_owner);
+    }
+
+    /// @notice send `_value` token to `_to` from `msg.sender`
+    /// @param _to The address of the recipient
+    /// @param _value The amount of token to be transferred
+    /// @return success Whether the transfer was successful or not
+    function transfer(address _to, uint256 _value) external returns (bool success) {
+        return _curve.transfer(_to, _value);
```
alcueca, line 24: This won't work, because `msg.sender` to `TokenProxy` is the user, and `msg.sender` to `_curve` is `TokenProxy`.

`return _curve.transferFrom(msg.sender, _to, _value)` might be what you need.

contracts/mocks/TokenProxy.sol
```diff
@@ -0,0 +1,54 @@
...
+    ) external returns (bool success) {
+        return _curve.transferFrom(_from, _to, _value);
+    }
+
+    /// @notice `msg.sender` approves `_addr` to spend `_value` tokens
+    /// @param _spender The address of the account able to transfer the tokens
+    /// @param _value The amount of wei to be approved for transfer
+    /// @return success Whether the approval was successful or not
+    function approve(address _spender, uint256 _value) external returns (bool success) {
+        return _curve.approve(_spender, _value);
```
alcueca, line 45: Same as above, `msg.sender` is different. Here I don't think that there is a workaround.

alcueca (changes requested):
I'm not sure what are you trying to do. Found a couple of errors, though.

contracts/mocks/TokenProxy.sol
iamsahu: I am trying to mock the `crv` & `cvx` contract in the `ConvexStakingWrapper` at the addresses mentioned in the [contract](https://github.com/yieldprotocol/vault-v2/blob/fix/convex-wrapper-mocking/contracts/utils/convex/ConvexStakingWrapper.sol#L37-L38).


I got this working by minting the tokens to `TokenProxy` [here](https://github.com/yieldprotocol/vault-v2/blob/fix/convex-wrapper-mocking/test/092_convex_wrapper.ts#L236-L237).

So, in theory the token is being transferred from `TokenProxy` & not the `ConvexPoolMock` which is the user calling the `transfer` but it is ok as we are just trying to simulate the rewards received from the `ConvexPool` as seen [here](https://github.com/yieldprotocol/vault-v2/blob/fix/convex-wrapper-mocking/contracts/utils/convex/ConvexStakingWrapper.sol#L283). So, it doesn't matter for the tests as to who actually is sending the rewards token as long as it is received after the `getReward` is called.

alcueca: All clear, thank you!

contracts/mocks/TokenProxy.sol
alcueca: As above

alcueca (approved):
LGTM

PR #376, opened 2022-02-15
ConvexYieldWrapper: Tight variable packing in `ConvexStakingWrapper.sol`

iamsahu:
Fix for the issue:
- https://github.com/code-423n4/2022-01-yield-findings/issues/58


alcueca (approved):
:+1:

PR #377, opened 2022-02-15
Convex Join

alcueca:
Now that we understand better how the ConvexStakingWrapper works, and we have been forced to refactor it heavily due to the poor code quality, we should reconsider whether we want to reuse the wrapping logic, but embed it with a Join instead of embedding it with an ERC20.

The wrapped convex token is not expected to have use outside the Yield Protocol, and it being transferrable poses security concerns that are solved by checkpointing on most transfers. Checkpoints between the wrapper and the original join are not checkpointed, though.

Merging the wrapping logic in the Join requires taking only `totalSupply` from ERC20, renamed to `managed_assets`. This counter increases and decreases in `join` and `exit`.

Most other changes are straightforward from here.

A second change that has been included in this PR (in a separate commit), is to make cvx a regular reward token. CVX gets added in position 0 of the rewards array. It gets checkpointed as any other reward, but when calculating earnings is loaded and calculated separately.

Another change (included in the first commit) has been to deduplicate the code between `_checkpoint` and `_checkpointAndClaim`.


contracts/utils/convex/ConvexJoin.sol
```diff
@@ -0,0 +1,359 @@
...
+
+        uint256 rewardsLength = rewards.length;
+
+        if (rewardsLength == 0) {
+            // We add cvx as a regular reward, but earnings will be calculated separately
+            RewardType storage reward = rewards.push();
+            reward.reward_token = cvx;
+            // There is no reward pool for cvx
+
+            RewardType storage reward = rewards.push();
```
iamsahu, line 177: Already declared on 173. Will throw an error.

alcueca:
Would you please:
Rename `Join.Sol` to `FlashJoin.sol`
Rename `BasicJoin.sol` to `Join.sol`
Make `FlashJoin.sol` inherit from `Join.sol` (removing all the repeated functionality, of course).

0xleastwood:
I think I'm happy with the changes. The only clarification I would want to make is how the `retreive()` function is intended to work?

alcueca:
Sorry for the back and forth, but the `checkpoint` needs to happen in `join`, taking into account that the ladle updates the vault first (so the cvx3CRV in the Join, about to be staked, needs to be removed from the aggregated balance to be checkpointed).

The `checkpoint` in exit can be removed completely, because at that point we don't even know the user that owned the cvx3CRV.

The frontend needs to be sure to call `checkpointAndClaim` as the first step of any batch that removes cvx3CRV from a vault, or the rewards will be lost.

alcueca:
@0xleastwood, the `retrieve` function is there to retrieve tokens that might be airdropped to the Join. We initially had COMP in mind. I'm not certain that we need to keep it in the basic Join, as it opens up an attack surface.

alcueca:
@iamsahu, please approve and merge this PR.

iamsahu (approved):

PR #378, opened 2022-02-16
Notional fCash as collateral

alcueca:
- Regular Join that takes ERC1155 assets, but is locked to a single id. No flash loans.
- Minimal Ladle module to transfer ERC1155 assets. No permit.
- Minimal Oracle that prices fCash at face value.

Check the README for more details.

All files have been stored in `contracts/other/notional`. We can have a discussion on how to organize this repo better.
1. I think that we could keep specific integration contracts in `other`, which would also mean moving a bunch of oracles.
2. The ERC1155.sol contract can be imported from solmate, or copied into `utils`.
3. The `Join1155` could go into a `joins` folder since it is not specific to any solution.

### DIFF
--- a/contracts/other/notional/ERC1155.sol
+++ b/contracts/other/notional/ERC1155.sol
@@ -1,0 +1,253 @@
+// SPDX-License-Identifier: AGPL-3.0-only
+pragma solidity >=0.8.0;
+
+/// @notice Minimalist and gas efficient standard ERC1155 implementation.
+/// @author Solmate (https://github.com/Rari-Capital/solmate/blob/main/src/tokens/ERC1155.sol)
+abstract contract ERC1155 {
+    /*///////////////////////////////////////////////////////////////
+                                EVENTS
+    //////////////////////////////////////////////////////////////*/
+
+    event TransferSingle(
+        address indexed operator,
+        address indexed from,
+        address indexed to,
+        uint256 id,
+        uint256 amount
+    );
+
+    event TransferBatch(
+        address indexed operator,
+        address indexed from,
+        address indexed to,
+        uint256[] ids,
+        uint256[] amounts
+    );
+
+    event ApprovalForAll(address indexed owner, address indexed operator, bool approved);
+
+    event URI(string value, uint256 indexed id);
+
+    /*///////////////////////////////////////////////////////////////
+                            ERC1155 STORAGE
+    //////////////////////////////////////////////////////////////*/
+
+    mapping(address => mapping(uint256 => uint256)) public balanceOf;
+
+    mapping(address => mapping(address => bool)) public isApprovedForAll;
+
+    /*///////////////////////////////////////////////////////////////
+                             METADATA LOGIC
+    //////////////////////////////////////////////////////////////*/
+
+    function uri(uint256 id) public view virtual returns (string memory);
+
+    /*///////////////////////////////////////////////////////////////
+                             ERC1155 LOGIC
+    //////////////////////////////////////////////////////////////*/
+
+    function setApprovalForAll(address operator, bool approved) public virtual {
+        isApprovedForAll[msg.sender][operator] = approved;
+
+        emit ApprovalForAll(msg.sender, operator, approved);
+    }
+
+    function safeTransferFrom(
+        address from,
+        address to,
+        uint256 id,
+        uint256 amount,
+        bytes memory data
+    ) public virtual {
+        require(msg.sender == from || isApprovedForAll[from][msg.sender], "NOT_AUTHORIZED");
+
+        balanceOf[from][id] -= amount;
+        balanceOf[to][id] += amount;
+
+        emit TransferSingle(msg.sender, from, to, id, amount);
+
+        require(
+            to.code.length == 0
+                ? to != address(0)
+                : ERC1155TokenReceiver(to).onERC1155Received(msg.sender, from, id, amount, data) ==
+                    ERC1155TokenReceiver.onERC1155Received.selector,
+            "UNSAFE_RECIPIENT"
+        );
+    }
+
+    function safeBatchTransferFrom(
+        address from,
+        address to,
+        uint256[] memory ids,
+        uint256[] memory amounts,
+        bytes memory data
+    ) public virtual {
+        uint256 idsLength = ids.length; // Saves MLOADs.
+
+        require(idsLength == amounts.length, "LENGTH_MISMATCH");
+
+        require(msg.sender == from || isApprovedForAll[from][msg.sender], "NOT_AUTHORIZED");
+
+        for (uint256 i = 0; i < idsLength; ) {
+            uint256 id = ids[i];
+            uint256 amount = amounts[i];
+
+            balanceOf[from][id] -= amount;
+            balanceOf[to][id] += amount;
+
+            // An array can't have a total length
+            // larger than the max uint256 value.
+            unchecked {
+                i++;
+            }
+        }
+
+        emit TransferBatch(msg.sender, from, to, ids, amounts);
+
+        require(
+            to.code.length == 0
+                ? to != address(0)
+                : ERC1155TokenReceiver(to).onERC1155BatchReceived(msg.sender, from, ids, amounts, data) ==
+                    ERC1155TokenReceiver.onERC1155BatchReceived.selector,
+            "UNSAFE_RECIPIENT"
+        );
+    }
+
+    function balanceOfBatch(address[] memory owners, uint256[] memory ids)
+        public
+        view
+        virtual
+        returns (uint256[] memory balances)
+    {
+        uint256 ownersLength = owners.length; // Saves MLOADs.
+
+        require(ownersLength == ids.length, "LENGTH_MISMATCH");
+
+        balances = new uint256[](owners.length);
+
+        // Unchecked because the only math done is incrementing
+        // the array index counter which cannot possibly overflow.
+        unchecked {
+            for (uint256 i = 0; i < ownersLength; i++) {
+                balances[i] = balanceOf[owners[i]][ids[i]];
+            }
+        }
+    }
+
+    /*///////////////////////////////////////////////////////////////
+                              ERC165 LOGIC
+    //////////////////////////////////////////////////////////////*/
+
+    function supportsInterface(bytes4 interfaceId) public pure virtual returns (bool) {
+        return
+            interfaceId == 0x01ffc9a7 || // ERC165 Interface ID for ERC165
+            interfaceId == 0xd9b67a26 || // ERC165 Interface ID for ERC1155
+            interfaceId == 0x0e89341c; // ERC165 Interface ID for ERC1155MetadataURI
+    }
+
+    /*///////////////////////////////////////////////////////////////
+                        INTERNAL MINT/BURN LOGIC
+    //////////////////////////////////////////////////////////////*/
+
+    function _mint(
+        address to,
+        uint256 id,
+        uint256 amount,
+        bytes memory data
+    ) internal {
+        balanceOf[to][id] += amount;
+
+        emit TransferSingle(msg.sender, address(0), to, id, amount);
+
+        require(
+            to.code.length == 0
+                ? to != address(0)
+                : ERC1155TokenReceiver(to).onERC1155Received(msg.sender, address(0), id, amount, data) ==
+                    ERC1155TokenReceiver.onERC1155Received.selector,
+            "UNSAFE_RECIPIENT"
+        );
+    }
+
+    function _batchMint(
+        address to,
+        uint256[] memory ids,
+        uint256[] memory amounts,
+        bytes memory data
+    ) internal {
+        uint256 idsLength = ids.length; // Saves MLOADs.
+
+        require(idsLength == amounts.length, "LENGTH_MISMATCH");
+
+        for (uint256 i = 0; i < idsLength; ) {
+            balanceOf[to][ids[i]] += amounts[i];
+
+            // An array can't have a total length
+            // larger than the max uint256 value.
+            unchecked {
+                i++;
+            }
+        }
+
+        emit TransferBatch(msg.sender, address(0), to, ids, amounts);
+
+        require(
+            to.code.length == 0
+                ? to != address(0)
+                : ERC1155TokenReceiver(to).onERC1155BatchReceived(msg.sender, address(0), ids, amounts, data) ==
+                    ERC1155TokenReceiver.onERC1155BatchReceived.selector,
+            "UNSAFE_RECIPIENT"
+        );
+    }
+
+    function _batchBurn(
+        address from,
+        uint256[] memory ids,
+        uint256[] memory amounts
+    ) internal {
+        uint256 idsLength = ids.length; // Saves MLOADs.
+
+        require(idsLength == amounts.length, "LENGTH_MISMATCH");
+
+        for (uint256 i = 0; i < idsLength; ) {
+            balanceOf[from][ids[i]] -= amounts[i];
+
+            // An array can't have a total length
+            // larger than the max uint256 value.
+            unchecked {
+                i++;
+            }
+        }
+
+        emit TransferBatch(msg.sender, from, address(0), ids, amounts);
+    }
+
+    function _burn(
+        address from,
+        uint256 id,
+        uint256 amount
+    ) internal {
+        balanceOf[from][id] -= amount;
+
+        emit TransferSingle(msg.sender, from, address(0), id, amount);
+    }
+}
+
+/// @notice A generic interface for a contract which properly accepts ERC1155 tokens.
+/// @author Solmate (https://github.com/Rari-Capital/solmate/blob/main/src/tokens/ERC1155.sol)
+interface ERC1155TokenReceiver {
+    function onERC1155Received(
+        address operator,
+        address from,
+        uint256 id,
+        uint256 amount,
+        bytes calldata data
+    ) external returns (bytes4);
+
+    function onERC1155BatchReceived(
+        address operator,
+        address from,
+        uint256[] calldata ids,
+        uint256[] calldata amounts,
+        bytes calldata data
+    ) external returns (bytes4);
+}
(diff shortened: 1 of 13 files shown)


alcueca:
Notional's fCash isn't an ERC1155 at all. It's not clear that it can be transferred at all. Even harder to assure that it can be transferred by someone else than the owner.

alcueca:
There is actually an ERC1155 wrapper for fCash

contracts/other/notional/ERC1155.sol
```diff
@@ -0,0 +1,253 @@
...
+            uint256 id = ids[i];
+            uint256 amount = amounts[i];
+
+            balanceOf[from][id] -= amount;
+            balanceOf[to][id] += amount;
+
+            // An array can't have a total length
+            // larger than the max uint256 value.
+            unchecked {
+                i++;
```
devtooligan, line 101:
```
++i;
```
😄

contracts/other/notional/Join1155.sol
```diff
@@ -0,0 +1,122 @@
+// SPDX-License-Identifier: BUSL-1.1
+pragma solidity 0.8.6;
+
+import "@yield-protocol/vault-interfaces/IJoin.sol";
+import "@yield-protocol/utils-v2/contracts/token/IERC20.sol";
+import "@yield-protocol/utils-v2/contracts/token/MinimalTransferHelper.sol";
+import "@yield-protocol/utils-v2/contracts/access/AccessControl.sol";
+import "@yield-protocol/utils-v2/contracts/math/WMul.sol";
+import "@yield-protocol/utils-v2/contracts/cast/CastU256U128.sol";
+import "./ERC1155.sol"; // TODO: Move to yield-utils-v2
```
devtooligan, line 10: do we want to do this before we release or nah?

contracts/other/notional/Join1155.sol
```diff
@@ -0,0 +1,122 @@
...
+
+    constructor(address asset_, uint256 id_) {
+        asset = asset_;
+        id = id_;
+    }
+
+    /// @dev Advertising through ERC165 the available functions
+    function supportsInterface(bytes4 interfaceID) external view returns (bool) {
+        return  interfaceID == Join1155.supportsInterface.selector ||    // ERC-165 support (i.e. `bytes4(keccak256('supportsInterface(bytes4)'))`).
+                interfaceID == ERC1155TokenReceiver.onERC1155Received.selector ^ ERC1155TokenReceiver.onERC1155BatchReceived.selector;      // ERC-1155 `ERC1155TokenReceiver` support (i.e. `bytes4(keccak256("onERC1155Received(address,address,uint256,uint256,bytes)")) ^ bytes4(keccak256("onERC1155BatchReceived(address,address,uint256[],uint256[],bytes)"))`).
```
devtooligan, line 35: this comment is a bit unwieldly can we move it to the next time or otherwise break it up?

contracts/other/notional/Join1155.sol
```diff
@@ -0,0 +1,122 @@
...
+    /// @dev Called by the sender after a batch transfer to verify it was received. Ensures only `id` tokens are received.
+    function onERC1155BatchReceived(
+        address _operator,
+        address _from,
+        uint256[] calldata _ids,
+        uint256[] calldata _values,
+        bytes calldata _data
+    ) external override returns(bytes4) {
+        uint256 length = _ids.length;
+        for (uint256 i; i < length; ++i)
```
devtooligan, line 59: 💯  nice loop
could replace ++i with a call to a new fn if we really wanted to go crazy -- 
```suggestion
        for (uint256 i; i < length; _incrementCounter(i);)
```
```
function _incrementCounter(uint256 index) internal pure returns(uint256) {
    unchecked {
        return index + 1;
    }
}
```
i believe it would be inlined by the compiler

contracts/other/notional/Join1155.sol
```diff
@@ -0,0 +1,122 @@
...
+// ERC1155TokenReceiver is in ERC1155.sol
+
+contract Join1155 is IJoin, ERC1155TokenReceiver, AccessControl() {
+    using WMul for uint256;
+    using CastU256U128 for uint256;
+
+    event FlashFeeFactorSet(uint256 indexed fee);
+
+    bytes32 constant internal FLASH_LOAN_RETURN = keccak256("ERC3156FlashBorrower.onFlashLoan");
+    uint256 constant FLASH_LOANS_DISABLED = type(uint256).max;
```
devtooligan, line 20: my lintere's complaining this doesn't have a visibility stated.  it's kinda weird to even have visibility on a constant as my understanding is that the value gets replaced during compile.  but we have visibility on the one above so at least I guess we should be consistent

contracts/other/notional/Join1155.sol
```diff
@@ -0,0 +1,122 @@
...
+    /// @dev Take `amount` `asset` from `user` using `transferFrom`, minus any unaccounted `asset` in this contract.
+    function _join(address user, uint128 amount)
+        internal
+        returns (uint128)
+    {
+        ERC1155 token = ERC1155(asset);
+        uint256 _storedBalance = storedBalance;
+        uint256 available = token.balanceOf(address(this), id) - _storedBalance; // Fine to panic if this underflows
+        storedBalance = _storedBalance + amount;
+        unchecked { if (available < amount) token.safeTransferFrom(user, address(this), id, amount - available, ""); }
```
devtooligan, line 82: we could prob include the line above in this checked block

contracts/other/notional/Join1155.sol
```diff
@@ -0,0 +1,122 @@
...
+    function retrieve(ERC1155 token, uint256 id_, address to)
+        external
+        auth
+    {
+        require(address(token) != address(asset) || id_ != id, "Use exit for asset");
+        token.safeTransferFrom(address(this), to, id_, token.balanceOf(address(this), id_), "");
+    }
+
+    /// @dev Retrieve any ERC20 tokens. Useful for airdropped tokens.
+    function retrieveERC20(IERC20 token, address to)
```
devtooligan, line 116: one idea we could consider to make this a little more consistent with regular join is we could also name this fn `retrieve` as well and it would choose the fn based on the args passed in.  If we do that not sure if we should switch order or `id` and `to` -- another idea would be call this one `retrieve` and rename the other one `retrieve1155`.   but either way just a nit as these fns are not even in the interface

contracts/other/notional/NotionalMultiOracle.sol
```diff
@@ -0,0 +1,81 @@
...
+
+    mapping(bytes6 => mapping(bytes6 => Source)) public sources;
+
+    /// @dev Set or reset an oracle source and its inverse
+    function setSource(bytes6 notionalId, bytes6 underlyingId, IERC20Metadata underlying)
+        external auth
+    {
+        require (notionalId != underlyingId, "Wrong input");
+        sources[notionalId][underlyingId] = Source({
+            baseDecimals: 18, // I'm assuming here that fCash has 18 decimals
```
devtooligan, line 33: is there no way to confirm this?

contracts/other/notional/NotionalMultiOracle.sol
```diff
@@ -0,0 +1,81 @@
...
+    }
+
+    /// @dev Convert amountBase base into quote at the latest oracle price.
+    function _peek(bytes6 baseId, bytes6 quoteId, uint256 amountBase)
+        private view
+        returns (uint amountQuote, uint updateTime)
+    {
+        Source memory source = sources[baseId][quoteId];
+        require (source.baseDecimals == 18 || source.quoteDecimals == 18, "Source not found"); // A bit meh as a test of existence
+        int price = 1e18; // We price fCash at face value
```
devtooligan, line 72: i don't know much about fcash but i see we are always valuing it at 1:1

is all this really necessary?  Can't we just `return (amount, block.timestamp)` always and call it a day?

contracts/other/notional/README.md
```diff
@@ -0,0 +1,35 @@
+# Notional fCash as collateral
+
+## About fCash
+Notional borrowing and lending positions are stored as [fCash](https://docs.notional.finance/notional-v2/notional-v2-basics/fcash).
+
+fCash positions can be positive or negative. If positive they are equivalent for a lending position, and are redeemable at maturity for their face value. If negative they are borrowing positions.
```
devtooligan, line 6: i guess I shouldve read this first lol

contracts/other/notional/README.md
```diff
@@ -0,0 +1,35 @@
...
+
+We want to use fCash lending positions as collateral. Given that we know that they are redeemable for their face value at maturity and their volatility is low, they are good assets for that purpose.
```
devtooligan, line 8: Are we fairly certain on this?  Should we consider if they are hacked or something and the face value cannot be redeemed?

contracts/other/notional/README.md
```diff
@@ -0,0 +1,35 @@
...
+
+Accepting fCash as collateral also eases arbitraging and leverage between Notional an Yield. By taking positive fCash balances as collateral, we facilitate lending on Notional and borrowing from Yield.
+
+fCash is kept internally in the Notional contracts, but there is a [proxy](https://github.com/notional-finance/contracts-v2/blob/master/contracts/external/actions/ERC1155Action.sol) that allows working with fCash as an ERC1155 token.
+
+The Notional [proxy](0x1344A36A1B56144C3Bc62E7757377D288fDE0369) is deployed on the Ethereum mainnet. It has links to the [ERC1155](0xffd7531ed937f703b269815950cb75bdaaa341c9) and [Views](0xde14d5f07456c86f070c108a04ae2fafdbd2a939) proxies. Not sure but I think that you call all functions on the main proxy.
```
devtooligan, line 14: Can we update this comment.  From what I saw we are only calling `safeTransferFrom` and when I go to the proxy address I don't see that fn exposed. https://etherscan.io/address/0x1344A36A1B56144C3Bc62E7757377D288fDE0369#writeProxyContract

test/020_oracles_notional.ts
```diff
@@ -0,0 +1,80 @@
...
+    expect(
+      (await notionalMultiOracle.callStatic.get(bytes6ToBytes32(FDAI), bytes6ToBytes32(DAI), WAD.mul(2500)))[0]
+    ).to.equal(WAD.mul(2500))
+    expect(
+      (await notionalMultiOracle.callStatic.get(bytes6ToBytes32(FUSDC), bytes6ToBytes32(USDC), WAD.mul(2500)))[0]
+    ).to.equal(oneUSDC.mul(2500))
+    expect(
+      (await notionalMultiOracle.callStatic.get(bytes6ToBytes32(DAI), bytes6ToBytes32(FDAI), WAD.mul(2500)))[0]
+    ).to.equal(WAD.mul(2500))
+    // expect((await notionalMultiOracle.callStatic.get(bytes6ToBytes32(USDC), bytes6ToBytes32(FUSDC), oneUSDC.mul(2500)))[0]).to.equal(
```
devtooligan, line 76: deleteme

test/020_oracles_notional.ts
```diff
@@ -0,0 +1,80 @@
...
+
+  it('retrieves the face value from a notional multioracle', async () => {
+    expect(
+      (await notionalMultiOracle.callStatic.get(bytes6ToBytes32(FDAI), bytes6ToBytes32(DAI), WAD.mul(2500)))[0]
+    ).to.equal(WAD.mul(2500))
+    expect(
+      (await notionalMultiOracle.callStatic.get(bytes6ToBytes32(FUSDC), bytes6ToBytes32(USDC), WAD.mul(2500)))[0]
+    ).to.equal(oneUSDC.mul(2500))
+    expect(
+      (await notionalMultiOracle.callStatic.get(bytes6ToBytes32(DAI), bytes6ToBytes32(FDAI), WAD.mul(2500)))[0]
```
devtooligan, line 74: should we throw a `peek()` in there for fun?

devtooligan:
Overall looks good.  I didn't find any dealbreakers, just some suggestions and a couple questions. 

Deploy scripts will be in a different pr, right?  And we will be doing some mainnet fork tests at that time.  Two things to track when doing the mainnet fork tests are:
1) Does the transfer work properly when pointed to the main proxy ctrct
2)  remember to do the `setApproval` for the ladle

I'm happy to approve this after a little more discussion on some of the questions I brought up, the actual code change suggestions are not that important to me

contracts/other/notional/NotionalMultiOracle.sol
aniemerg: We can value fCash at 1:1 because we are still using a collateralization ratio/loan-to-value. I believe the loan-to-value is 90%, meaning that the position is still above water even if interest rates hit 20% for a six-month term. Even if interest rates go above that, Yield Protocol doesn't lose money as it could always redeem the fCash for the underlying asset itself, even if the user never repays the loan.

devtooligan: I see.  What is different about fcash that we can use 1, but Dai or USDC we have to check prices?  We also use a collateralization ratio for those.

contracts/other/notional/Join1155.sol
aniemerg: should this be a constant? What if we want to turn flashloans on?

contracts/other/notional/NotionalMultiOracle.sol
aniemerg: fCash Dai settles to Dai at 1:1. So, we can value fCash Dai as being equal to Dai, with the only risk being that if we are wrong at that time, we may have to wait until maturity for it to be true. This assumption only works well if the collateral has the same (or an earlier) maturity than what is borrowed.

aniemerg (approved):

iamsahu (approved):

contracts/other/notional/Join1155.sol
alcueca: I'm a bit conflicted. Do we really want to copy ERC1155 from solmate to yield-utils-v2? Or we import solmate? Do we then deprecate yield-utils-v2 ERC20.sol and use solmate's instead?

contracts/other/notional/Join1155.sol
alcueca: Fixed

contracts/other/notional/Join1155.sol
alcueca: Get out of here, you punk. I like my code clean.

contracts/other/notional/Join1155.sol
alcueca: It's a valid concern. If no visibility is declared then `internal` is assumed. Once deployed, it would give some background as to why the flash loan fee is what it is.

alcueca: Fixed

contracts/other/notional/Join1155.sol
alcueca: Done, including comment as to why this is safe

contracts/other/notional/Join1155.sol
alcueca: I took the second option, as overloading is still quite clunky with ethers.

contracts/other/notional/NotionalMultiOracle.sol
alcueca: In my TODO for this morning

contracts/other/notional/README.md
alcueca: image

contracts/other/notional/README.md
alcueca: There are more proxies behind the main proxy, which expose that function.

image

test/020_oracles_notional.ts
alcueca: fixed

test/020_oracles_notional.ts
alcueca: Done

contracts/other/notional/README.md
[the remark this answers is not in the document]
alcueca: Done

alcueca:
No one of you found [the brown m&m](https://github.com/yieldprotocol/vault-v2/pull/378/commits/b353a86fb9df7905231498cc8e48f13369ce1eb9). Tsk, tsk.

contracts/other/notional/Join1155.sol
devtooligan: Implementing Solmate ERC20 into the `yield-utils-v2` repo sounds like something to consider but outside of scope for this pr.  As for the 1155 contract, I'm cool with doing it this way (since this is kind of a one-off contract anyways) and just removing the comment here. 

We could create a new ticket in the `yield-utils-v2` repo "Use Solmate" and next time we have to interact with a ERC20 or ERC1155 contract, we can think about the best way to do that?

Lemme know if you want me to make that ticket ^^

contracts/other/notional/README.md
devtooligan: Great.  Can we remove the "Not sure but I think..." from the README?

contracts/other/notional/NotionalMultiOracle.sol
devtooligan: @alcueca I understand better the thinking behinding valuing these at 1, but i still don't understand why we need all these steps if it's always going to return amount in the end.  It's a waste of gas and confusing. 

> Can't we just return (amount, block.timestamp) always and call it a day?

contracts/other/notional/NotionalMultiOracle.sol
alcueca: image

alcueca: Upon testing it's 8 decimals for everything.

contracts/other/notional/NotionalMultiOracle.sol
alcueca: Because fCash has 8 decimals, the base could have 18 (DAI) or 6 (USDC), so we need to convert from 1000000 USDC to 100000000 fUSDC, or 100000000 fDAI to 1000000000000000000 DAI.

alcueca: Simplified the code a bit, tho.

contracts/other/notional/README.md
alcueca: Done. Also tested that yesterday.

alcueca:
I'm merging this PR now, thanks for the feedback received. It remains a bit of thinking on how to reorganize the repository for continued development.

devtooligan (approved):

contracts/other/notional/Join1155.sol
devtooligan: done https://github.com/yieldprotocol/yield-utils-v2/issues/33

PR #381, opened 2022-03-01
Contango contracts

bbonanno:


contracts/contango/ContangoCauldron.sol
```diff
@@ -0,0 +1,115 @@
...
+    constructor(
+        uint128 _collateralisationRatio,
+        bytes6 _commonCurrency,
+        uint8 _commonCcyDecimals
+    ) {
+        collateralisationRatio = _collateralisationRatio * uint128(10**(18 - _commonCcyDecimals));
+        commonCcy = _commonCurrency;
+    }
+
+    function setParams(
```
alcueca, line 33: I'd suggest splitting this function into `setRatio` and `setCurrency`.

contracts/contango/ContangoCauldron.sol
```diff
@@ -0,0 +1,115 @@
...
+
+    uint128 public collateralisationRatio;
+    bytes6 public commonCcy; // Currency to use as common ground for all the ink & art
+
+    constructor(
+        uint128 _collateralisationRatio,
+        bytes6 _commonCurrency,
+        uint8 _commonCcyDecimals
+    ) {
+        collateralisationRatio = _collateralisationRatio * uint128(10**(18 - _commonCcyDecimals));
```
alcueca, line 29: Why not just passing the ratio?

contracts/contango/ContangoCauldron.sol
```diff
@@ -0,0 +1,115 @@
...
+
+contract ContangoCauldron is Cauldron {
+    using CauldronMath for uint128;
+    using EnumerableSet for EnumerableSet.Bytes6Set;
+    using CastU256I256 for uint256;
+    using WMul for uint256;
+    using WDivUp for uint256;
+
+    EnumerableSet.Bytes6Set private assetsInUse;
+    mapping(bytes6 => DataTypes.Balances) public balancesPerAsset;
```
alcueca, line 18: Are we doing away with individual vaults completely, or having a protocol vault and individual vaults in parallel?

contracts/contango/ContangoCauldron.sol
```diff
@@ -0,0 +1,115 @@
...
+        ) = vaultData(vaultId, true);
+
+        balances_ = _pour(vaultId, vault_, balances_, series_, ink, art);
+
+        _pour(vault_.ilkId, ink, vault_.seriesId, art);
+
+        return balances_;
+    }
+
+    function _pour(
```
alcueca, line 60: I would do:
```
        bytes6 inkId,
        bytes6 artId,
        int128 ink,
        int128 art
```
All through the code, I've put identifiers first, amounts afterwards.

alcueca (commented):
Mostly naming things, and one serious question about individual vaults.

contracts/contango/ContangoCauldron.sol
bbonanno: We have both, this impl calls the `_pour` on the parent that does all the existent stuff, then it replaces the call to `_level` by its internal call to `_pour` to update the global values and do global collateralisation check.
Maybe we could rename the second `_pour` to something else?

contracts/contango/ContangoCauldron.sol
bbonanno: ok

contracts/contango/ContangoCauldron.sol
bbonanno: no reason, started as storing the 3 things and ended up like this, will change it

contracts/contango/ContangoCauldron.sol
bbonanno: ok

contracts/contango/ContangoCauldron.sol
```diff
@@ -0,0 +1,115 @@
...
+        bytes6 artId,
+        int128 art
+    ) internal {
+        if (ink != 0) {
+            assetsInUse.add(inkId);
+            balancesPerAsset[inkId].ink = balancesPerAsset[inkId].ink.add(ink);
+        }
+        if (art != 0) {
+            assetsInUse.add(artId);
+            balancesPerAsset[artId].art = balancesPerAsset[artId].art.add(art);
```
alcueca, line 67: It is a bit weird that all assets and series can be both collateral and debt. It's ok to have fyUSDC as collateral, but here you can have USDC as debt as well. You only call this from `pour` so you know that you are not doing that, but from a data integrity perspective it looks weak to me.

contracts/contango/ContangoLadle.sol
```diff
@@ -0,0 +1,19 @@
...
+
+    constructor (ICauldron cauldron, IWETH9 weth) Ladle(cauldron, weth) { }
+
+    // @dev auth version of build, only contango can create vaults here
+    // all other methods rely on being vault owner, so no need to secure them
+    function build(bytes6 seriesId, bytes6 ilkId, uint8 salt)
+        external override payable auth
+        returns(bytes12, DataTypes.Vault memory)
+    {
+        return _build(seriesId, ilkId, salt);
```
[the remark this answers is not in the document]
alcueca: Now that `build` is `auth` they can be deterministic, no risk of DoS.

alcueca (changes requested):
I think that we need to think again on the protocol vault, and code it more clearly.

contracts/contango/ContangoLadle.sol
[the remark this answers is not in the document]
bbonanno: So I guess hardcoding a 0 to the internal call is what we want?

contracts/contango/ContangoCauldron.sol
bbonanno: All right, I'll rework that

contracts/contango/ContangoCauldron.sol
```diff
@@ -0,0 +1,110 @@
...
+        require(getFreeCollateralUSD() >= 0, "Undercollateralised");
+    }
+
+    function getFreeCollateralUSD() public returns (int256) {
+        (uint128 _collateralisationRatio, bytes6 _commonCurrency) = (collateralisationRatio, commonCcy);
+
+        uint256 totalInk;
+        uint256 totalArt;
+
+        //TODO is this reading .length() from storage each time?
```
alcueca, line 79: As far as I understand, it is, @devtooligan?

contracts/contango/ContangoCauldron.sol
```diff
@@ -0,0 +1,110 @@
...
+        for (uint256 index; index < assetsInUse.length(); index++) {
+            bytes6 assetId = assetsInUse.get(index);
+            DataTypes.Balances memory _balances = balancesPerAsset[assetId];
+            if (assetId != _commonCurrency && (_balances.ink > 0 || _balances.art > 0)) {
```
alcueca, line 83: You can rearrange this code:
```
if balances.art > 0
    totalArt += commonCurrency ? foo : bar
if balances.ink > 0 ...
```

contracts/contango/ContangoCauldron.sol
```diff
@@ -0,0 +1,110 @@
...
+                IOracle oracle = spotOracles[assetId][_commonCurrency].oracle;
+                totalInk += _valueAsset(oracle, assetId, _commonCurrency, _balances.ink);
+                totalArt += _valueAsset(oracle, assetId, _commonCurrency, _balances.art);
+            } else {
+                //TODO maybe remove unsed asset?
```
alcueca, line 88: Unused assets are ignored, so you can make that kind of garbage collection function an external one, called only when desired.

bbonanno: yeah, was torn in between automatic vs manual :)
I guess manual it is then

contracts/contango/ContangoCauldron.sol
devtooligan: yes should prob cash the length to save gas.  

while we're at i think we can just cash the whole `assetsInUse` to save gas on read in line 81?

and can also do something about index++ --> ideally make the increment unchecked or at least use ++index instead

bbonanno: ok, I'll do a pass on gas optimisations after I have the unit tests

bbonanno: Can't cache the whole `assetsInUse` as it contains nested mappings, but I addressed the increment and cached the length

alcueca (approved):
Let's merge, great work.

PR #382, opened 2022-03-09
Mature fcash

alcueca:
FCash can't be transferred after maturity, so we are forced to have the Join redeem all fCash holdings after maturity, and become an underlying exit-only Join.


contracts/other/notional/Join1155.sol
```diff
@@ -88,34 +120,76 @@ contract Join1155 is IJoin, ERC1155TokenReceiver, AccessControl() {
...
+            depositActionAmount: 0,
+            withdrawAmountInternalPrecision: 0,
+            withdrawEntireCashBalance: true,
+            redeemToUnderlying: true
+        });
+
+        IBatchAction(asset).batchBalanceAction(address(this), withdrawActions);
+
+        uint256 underlyingBalance = IERC20(underlying).balanceOf(address(this));
+        accrual = underlyingBalance.wdiv(storedBalance); // There is a rounding loss here. Some wei will be forever locked in the join.
```
aniemerg, line 179: Yikes! How will this affect users?

alcueca: As discussed, the rounding will not affect users. The underlying received cannot be evenly divided between holders with a set number of decimals, so we round down on `accrual` and up to one wei per user will remain in the contract.

PR #389, opened 2022-03-25
Migrate Hardhat test to Foundry for Chainlink Multioracle

Sabnock01:
This draft PR is for the migration of the vault-v2 tests in Hardhat to Foundry. 

I've only done the `ChainlinkMultiOracle` test thus far before getting feedback but I'm happy to add as many as needed to this. 

I use the node module style remappings so that we can still test in both Hardhat and Foundry as needed. 

The `Test.sol` file is a combination of [ds-test](https://github.com/dapphub/ds-test) and the [Vm interface](https://github.com/brockelmore/forge-std/blob/master/src/Vm.sol) in [forge-std](https://github.com/brockelmore/forge-std)

Other than that I follow the tests as closely as possible to those in Hardhat with `TestContants.sol` almost directly equivalent to `src/constants.ts` Will update the file as needed. Any constants declared locally in a test I do the same. 

What do we think of this architecture?

Any feedback or corrections are welcome.


devtooligan (approved):
nice work.  @alcueca merging this should be fine, It won't cause any probs with existing hh tests

devtooligan:
@Sabnock01 I'd be down with merging this one first as a kind of tester to prove that adding these tests won't mess anything up with the repo

Sabnock01:
> @Sabnock01 I'd be down with merging this one first as a kind of tester to prove that adding these tests won't mess anything up with the repo

I think that's a good idea.

alcueca (approved):
LGTM

PR #391, opened 2022-04-02
Foundry tests

Sabnock01:
This PR is for the conversion of the CompoundMultiOracle test to Foundry.


contracts/foundry-tests/ChainlinkMultiOracle.t.sol
```diff
@@ -0,0 +1,71 @@
...
+        chainlinkMultiOracle = new ChainlinkMultiOracle();
+        chainlinkMultiOracle.grantRole(0xef532f2e, address(this));
+        chainlinkMultiOracle.setSource(DAI, dai, ETH, weth, address(daiEthAggregator));
+        chainlinkMultiOracle.setSource(USDC, usdc, ETH, weth, address(usdcEthAggregator));
+        vm.warp(uint256(mockBytes32));
+        daiEthAggregator.set(WAD / 2500);
+        usdcEthAggregator.set(WAD / 2500);
+    }
+
+    function testGetSpotPrice() public {
```
alcueca, line 43: Actually, this doesn't test the spotPrice. This should be `testGetConversion`. We are testing the value of `base` in `quote` terms using `get` (we could be using `peek` as well).

Below it's the same in all tests. We test an asset conversion, using `get` or `peek`.

[Related reading](https://hackernoon.com/getting-prices-right)

contracts/foundry-tests/RateChiMultioracle.t.sol
```diff
@@ -0,0 +1,45 @@
...
+    }
+
+    function testRevertUnknownSource() public {
+        vm.expectRevert("Source not found");
+        compoundMultiOracle.get(mockBytes6, mockBytes6, WAD);
+    }
+
+    function testSetRetrieveChiRate() public {
+        cTokenChi.set(WAD * 2);
+        cTokenRate.set(WAD * 3);
```
alcueca, line 35: The two `set` commands should be called in `setUp` instead, and the test should be `testRetrieveChiRate`

alcueca (changes requested):
Some corrections on test naming, some reading as background, and a request for fuzzing tests.

contracts/foundry-tests/RateChiMultioracle.t.sol
Sabnock01: The TypeScript test has them both in the `it()` block so I thought I'd follow it but I like having them in the `setUP` too and this will make it uniform with the Chainlink multioracle test as well.

contracts/foundry-tests/RateChiMultioracle.t.sol
```diff
@@ -0,0 +1,45 @@
...
+    }
+
+    function testRevertUnknownSource() public {
+        vm.expectRevert("Source not found");
+        compoundMultiOracle.get(mockBytes6, mockBytes6, WAD);
+    }
+
+    function testSetRetrieveChiRate() public {
```
alcueca, line 35: `chi` is just that, not `chiPrice`. Same for `rate`.

`chi` is the factor by which savings increase, and `rate` is the factor by which debts increase.

contracts/foundry-tests/LidoOracle.t.sol
```diff
@@ -0,0 +1,106 @@
...
+        require(wstethStethConversion == 1008339308050006006);
+    }
+
+    function testRevertOnUnknownSource() public {
+        vm.expectRevert("Source not found");
+        lidoOracle.get(bytes32(DAI), bytes32(mockBytes6), WAD);
+    }
+
+    function testRetrieveDirectPairConversion() public view {
+        (uint256 wstethStethConversion,) = compositeMultiOracle.peek(WSTETH, STETH, 1e18);
```
alcueca, line 77: I find it a lot clearer to use ticker+`Amount` for variables.

```
uint256 wstEthAmount = 1e18;
(uint256 stEthAmount,) = compositeMultiOracle.peek(WSTETH, STETH, wstEthAmount);
```

Same in all subsequent tests.

contracts/foundry-tests/LidoOracle.t.sol
```diff
@@ -0,0 +1,106 @@
...
+    ChainlinkMultiOracle public chainlinkMultiOracle;
+    CompositeMultiOracle public compositeMultiOracle;
+    ChainlinkAggregatorV3Mock public stethEthAggregator;
+    ChainlinkAggregatorV3Mock public usdcEthAggregator;
+
+    bytes6 public mockBytes6 = 0xd1eaa762fae7;
+
+    function setUp() public {
+        lidoMock = new WstETHMock();
+        lidoMock.set(1008339308050006006);
```
alcueca, line 31: These magic numbers should be extracted to variables.

```
uint256 stEthToWstEthPrice = 1008339308050006006
```

A comment should be also added to remind reviewers that `stEthToWstEthPrice` is the amount of wstEth that you get for 1e18 stEth.

contracts/foundry-tests/LidoOracle.t.sol
```diff
@@ -0,0 +1,106 @@
...
+        compositeMultiOracle.setPath(WSTETH, ETH, path);
+        bytes6[] memory paths = new bytes6[](2);
+        paths[0] = STETH;
+        paths[1] = ETH;
+        compositeMultiOracle.setPath(WSTETH, USDC, paths);
+    }
+
+    function testGetConversion() public {
+        (uint256 stethWstethConversion,) = lidoOracle.get(STETH, WSTETH, WAD);
+        require(stethWstethConversion == 991729660855795538);
```
alcueca, line 66: Once we use the price variables, this require can be rewritten as:
```
require(stethWstethConversion == stEthAmount.muld(stEthToWstEthPrice));
```

Note that you need to import the fixed point math libraries.

Same for all instances below.

alcueca (changes requested):
Let's set a high bar on how tests are written, so that we obtain a test suite that will last us for a long time. Requested some variable name changes, and using variables to store magic numbers.

contracts/foundry-tests/RateChiMultioracle.t.sol
Sabnock01: Understood. Hopefully `getChi`/`peekChi` and `getRate`/`peekRate` will be sufficient for the differentiation.

contracts/foundry-tests/LidoOracle.t.sol
bbonanno: The right way to do this is using assertEq or any of the other assert* methods coming from Test/DSTest. I've updated the Chainlink test that's on master to reflect that, check #400

Sabnock01: I can refactor the tests to use `assertEq` in place of `require` but what was your reasoning for doing so precisely?

alcueca (approved):
Thanks for doing this :+1:

PR #393, opened 2022-04-06
publish all contracts but the mocks

bbonanno:


alcueca (approved):
Thanks for the clean up, and published.

PR #397, opened 2022-04-20
Rename LidoWrapHandler

alcueca:
Using a name pattern as in "Unwrapped token ticker"+"Converter".
Removing unneeded dependency on LadleStorage
Removing events as superfluous
Renaming variables to "token ticker"+"Amount"
Publishing to npm


contracts/other/lido/StEthConverter.sol
```diff
@@ -0,0 +1,37 @@
...
+    constructor(IWstETH wstETH_, IERC20 stETH_) {
+        wstETH = wstETH_;
+        stETH = stETH_;
+        stETH_.approve(address(wstETH_), type(uint256).max);
+    }
+
+    /// @dev Wrap stEth held by this contract and forward it to the "to" address
+    function wrap(address to) external returns (uint256 wstEthAmount) {
+        uint256 stEthAmount = stETH.balanceOf(address(this));
+        require(stEthAmount > 0, "No stETH to wrap");
```
devtooligan, line 25: this check is redundant shall we remove it?
Screen Shot 2022-04-20 at 8 27 16 AM
https://etherscan.io/address/0x7f39c581f595b53c5cb19bd0b3f8da6c935e2ca0#code

contracts/other/lido/StEthConverter.sol
```diff
@@ -0,0 +1,37 @@
...
+        wstEthAmount = wstETH.wrap(stEthAmount);
+        wstETH.safeTransfer(to, wstEthAmount);
+    }
+
+    /// @dev Unwrap WstETH held by this contract, and send the stETH to the "to" address
+    function unwrap(address to) external returns (uint256 unwstEthAmount) {
+        uint256 wstEthAmount = wstETH.balanceOf(address(this));
+        require(wstEthAmount > 0, "No wstETH to unwrap");
```
devtooligan, line 33: same

devtooligan (approved):
All the new changes look great, love the new naming convention.  I noticed a couple redundant zero checks in there - not introduced by this pr but shall we remove them?

contracts/other/lido/StEthConverter.sol
alcueca: Lovely

devtooligan:
I guess we need to update the wstEthMock to revert on shares == 0

alcueca:
Tbh, I don't care about users executing zero amount transactions.

PR #400, opened 2022-05-05
Migrate Witch tests to foundry

ultrasecreth:
- update foundry deps
- update foundry test
- forge install: forge-std
- use forge-std
- clean existing test
- Witch auction tests
- All witch tests ported to foundry
- Foundry tests added to CI

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -4,6 +4,3 @@
 [submodule "lib/yield-utils-v2"]
 	path = lib/yield-utils-v2
 	url = https://github.com/yieldprotocol/yield-utils-v2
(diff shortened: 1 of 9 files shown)


Sabnock01:
In what way is `forge-std` not HH compatible?

bbonanno:
> In what way is `forge-std` not HH compatible?

The project is not on npm.
It can't be added as a git dependency as it doesn't have a package.Json

That means hardhat can't see it so it can't compile the foundry-tests folder.

In my experience, rather than having all together, it's better to use npm workspaces to make a clear separation of hardhat and foundry

PR #403, opened 2022-05-11
Giver contract

iamsahu:
A simple contract that allows owner of a vault to give the vault to another address. In the proposal [ypp-0036](https://github.com/yieldprotocol/governance-v2/pull/36), the give permission for ladle was removed.
For `YieldLever` & similar contracts in the future we would need a way to let users give the vault to another address for different reasons, for eg: flash loan. This contract will safely allow the give function to be accessed by the intended users.


contracts/utils/Giver.sol
```diff
@@ -0,0 +1,38 @@
...
+pragma solidity 0.8.6;
+
+import "@yield-protocol/vault-interfaces/src/ICauldron.sol";
+import "@yield-protocol/vault-interfaces/src/DataTypes.sol";
+import "@yield-protocol/utils-v2/contracts/access/AccessControl.sol";
+
+/// @title A contract that allows owner of a vault to give the vault
+contract Giver is AccessControl {
+    mapping(bytes6 => bool) public ilkBlacklist;
+    ICauldron public immutable cauldron;
```
alcueca, line 11: For code style, please put immutables and constants at the top of the state variable section.

contracts/utils/Giver.sol
```diff
@@ -0,0 +1,38 @@
...
+    /// @param ilkId Ilkid to be blacklisted
+    event IlkBlacklisted(bytes6 ilkId);
+
+    constructor(ICauldron cauldron_) {
+        cauldron = cauldron_;
+    }
+
+    /// @notice Function to blacklist
+    /// @param ilkId the ilkId to be blacklisted
+    function blacklistIlk(bytes6 ilkId) external auth {
```
alcueca, line 23: Please make this a `function banIlk(bytes6 ilkId, bool set) external auth`, and `set` is used to ban/unban ilks.

Let's use `ban` instead of `blacklist`, short and clear.

alcueca (approved):
Please add the tests before merging.

PR #411, opened 2022-05-17
Witch refactor

alcueca:
Branched out from [this commit](https://github.com/yieldprotocol/vault-v2/pull/365/commits/1f15b965f19f77dddf5fc74aae2bc97968655d18).

Tests for the Witch have been completely disabled, and will be rewritten in Foundry in subsequent PRs.


contracts/Witch.sol
```diff
@@ -13,198 +13,358 @@ import "@yield-protocol/utils-v2/contracts/math/WDivUp.sol";
...
     /// @dev Point to a different ladle
+    /// @param param Name of parameter to set (must be "ladle")
+    /// @param value Address of new ladle
     function point(bytes32 param, address value) external auth {
-        if (param == "ladle") ladle = ILadle(value);
-        else revert("Unrecognized parameter");
+        require(param == "ladle", "Unrecognized");
+        ladle = ILadle(value);
         emit Point(param, value);
     }
```
bbonanno, line 75: any reason for a generic method that can only set the ladle instead of something like `setLadle(address)`?

contracts/Witch.sol
```diff
@@ -13,198 +13,358 @@ import "@yield-protocol/utils-v2/contracts/math/WDivUp.sol";
...
+    /// @param initialOffer Proportion of collateral that is sold at auction start (1e18 = 100%)
+    function setLine(
+        bytes6 ilkId,
+        bytes6 baseId,
+        uint32 duration,
+        uint64 proportion,
+        uint64 initialOffer
+    ) external auth {
+        require(initialOffer <= 1e18, "InitialOffer above 100%");
+        require(proportion <= 1e18, "Proportion above 100%");
```
bbonanno, line 91: I'd add `require(initialOffer >= 0.01e18)` or something like that to mitigate the chance of repeating the issue of config values with the wrong precision. Same for `proportion`

contracts/Witch.sol
```diff
@@ -13,198 +13,358 @@ import "@yield-protocol/utils-v2/contracts/math/WDivUp.sol";
...
     using CastU256U128 for uint256;
     using CastU256U32 for uint256;
 
-    event Point(bytes32 indexed param, address value);
-    event IlkSet(bytes6 indexed ilkId, uint32 duration, uint64 initialOffer, uint96 line, uint24 dust, uint8 dec);
-    event Bought(bytes12 indexed vaultId, address indexed buyer, uint256 ink, uint256 art);
     event Auctioned(bytes12 indexed vaultId, uint256 indexed start);
+    event Bought(bytes12 indexed vaultId, address indexed buyer, uint256 ink, uint256 art);
+    event LineSet(bytes6 indexed ilkId, bytes6 baseId, uint32 duration, uint64 proportion, uint64 initialOffer);
+    event LimitSet(bytes6 indexed ilkId, bytes6 baseId, uint96 max, uint24 dust, uint8 dec);
```
bbonanno, line 33: Maybe is just me, but I find it easier to think on terms of base -> collateral, so I'd be tempted to flip the first 2 values. Also I think the base should be indexed too

contracts/Witch.sol
```diff
@@ -13,198 +13,358 @@ import "@yield-protocol/utils-v2/contracts/math/WDivUp.sol";
...
-            delete auctions[vaultId];
+        DataTypes.Vault memory vault = cauldron.vaults(vaultId);
+        DataTypes.Series memory series = cauldron.series(vault.seriesId);
+
+        // Find out how much debt is being repaid
+        uint128 artIn = uint128(cauldron.debtFromBase(vault.seriesId, maxBaseIn));
+
+        // If offering too much base, take only the necessary.
+        artIn = artIn > auction_.art ? auction_.art : artIn;
+        baseIn = cauldron.debtToBase(vault.seriesId, auction_.art);
```
bbonanno, line 193: shouldn't `artIn` be used in the second param? The guy may be paying for half of the `auction.art`

contracts/Witch.sol
```diff
@@ -13,198 +13,358 @@ import "@yield-protocol/utils-v2/contracts/math/WDivUp.sol";
...
-        {
-            uint256 elapsed = uint32(block.timestamp) - auction_.start;                      // Auctions will malfunction on the 7th of February 2106, at 06:28:16 GMT, we should replace this contract before then.
-            uint256 price = inkPrice(balances_, ilk_.initialOffer, ilk_.duration, elapsed);
-            ink = uint256(art).wmul(price);                                                    // Calculate collateral to sell. Using divdrup stops rounding from leaving 1 stray wei in vaults.
-            require (ink >= min, "Not enough bought");
-            require (art == balances_.art || balances_.ink - ink >= limits_.dust * (10 ** limits_.dec), "Leaves dust");
-            limits[vault_.ilkId].sum -= ink.u128();
-        }
+    /// @dev Pay `base` of the debt in a vault in liquidation, getting at least `minInkOut` collateral.
+    /// Use `payAll` to pay all the debt, using `paySome` for amounts close to the whole vault might revert.
```
bbonanno, line 166: `payAll` doesn't exist anymore, I think the msg should be similar to the `payWithTokens` one, so the liquidator just gives enough allowance to buy all knowing he won't overpay

contracts/Witch.sol
alcueca: Because all across the protocol we use that same method for reorchestration.

contracts/Witch.sol
alcueca: Fixed, thanks. Allowed 0% as well for `initialProportion`

contracts/Witch.sol
alcueca: True, fixed. Thanks.

contracts/Witch.sol
alcueca: Me too, but all across the protocol is ink first, art/base afterwards. We took that from MakerDAO in v1, and we have kept with it since. Better not to change it now, unfortunately.

contracts/Witch.sol
```diff
@@ -7,204 +7,395 @@ import "@yield-protocol/vault-interfaces/src/ICauldron.sol";
...
+    function setLine(
+        bytes6 ilkId,
+        bytes6 baseId,
+        uint32 duration,
+        uint64 proportion,
+        uint64 initialOffer
+    ) external auth {
+        require(initialOffer <= 1e18, "InitialOffer above 100%");
+        require(proportion <= 1e18, "Proportion above 100%");
+        require(initialOffer == 0 || initialOffer >= 1e16, "InitialOffer below 1%");
```
bbonanno, line 87: 0.01e18 is valid and more readable imho :)

bbonanno (changes requested):

bbonanno:
I'm writing the bot contracts against this new `Witch`, so would be nice to keep this PR as up-to-date as possible :)

bbonanno:
I just realised it'd be quite useful to have a couple of `view` functions:

- One that can tell me how much `ink` and `art` would be put for auction for a given `vaultId` (maybe even if the vault is not ready to be liquidated yet).
- One that can quote how much `ink` I'd get by repaying X amount of `art` (regardless if the vault has been auctioned already or not).

This is to avoid duplication (and fundamentally issues) on the liquidation bots, being able to approximate how much I'd get right now lets me know if I can `auction` & `pay` on the same tx, or if I should just `auction` and let time pass. And for the latter, being able to know how much I'd get right now helps me decide if it's time to `pay` or not yet.

bbonanno:
Replaced by #434

PR #419, opened 2022-05-23
YieldSpaceMultiOracle

bbonanno:
- forge install: yieldspace-interfaces
- ported generic YieldSpaceMultiOracle from Contango
- rename sut
- forge install: yieldspace-v2
- Add YieldSpaceMultiOracle

### DIFF
--- a/contracts/foundry-tests/oracles/PoolOracle.t.sol
+++ b/contracts/foundry-tests/oracles/PoolOracle.t.sol
@@ -129,9 +129,25 @@ contract PoolOracleTest is Test {
         address pool = Mocks.mock("IPool");
 
         // Saturday, 26 February 2022 10:15:28
-        vm.warp(1645870528);
-        IPool(pool).cumulativeBalancesRatio.mock(6081248333687398745122493757731327);
-        IPool(pool).getCache.mock(1102997721918618810252394, 1199964191067551228067732, 1645859234);
+        uint256 timestamp = 1645870528;
+        vm.warp(timestamp);
+        _stubPoolState(
+            // cast find-block 1645870528 # Saturday, 26 February 2022 10:15:28
+            // 14281108
+            // cast call 0x5D14Ab14adB3a3D9769a67a1D09634634bdE4C9B "cumulativeBalancesRatio()(uint256)" -b 14281108
+            // 6081248333687398745122493757731327
+            // cast call 0x5D14Ab14adB3a3D9769a67a1D09634634bdE4C9B "getCache()(uint112,uint112,uint32)"  -b 14281108
+            // 1102997721918618810252394
+            // 1199964191067551228067732
+            // 1645859234
+            PoolState({
+                pool: pool,
+                cumulativeBalancesRatio: 6081248333687398745122493757731327,
+                baseCached: 1102997721918618810252394,
+                fyTokenCached: 1199964191067551228067732,
+                blockTimestampLast: 1645859234
+            })
+        );
         oracle.update(pool);
 
         vm.record();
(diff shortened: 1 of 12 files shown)


contracts/oracles/yieldspace/YieldSpaceMultiOracle.sol
```diff
@@ -0,0 +1,132 @@
...
+        int128 timeTillMaturity = uint128(source.maturity - updateTime).fromUInt();
+        int128 powerValue64 = source.g.mul(source.ts).mul(timeTillMaturity);
+        // scale to 18 dec and convert to regular non-64
+        uint128 powerValue = uint128(powerValue64.mul(wad64x64).toUInt());
+
+        uint256 top = unitPrice.pow(powerValue, uint128(1e18));
+        uint256 bottom = uint128(1e18).pow(powerValue, uint128(1e18)) / 1e18;
+        uint256 marginalPrice = top / bottom;
+
+        if (source.inverse) {
```
alcueca, line 126:
```suggestion
        discountedAmount = source.inverse ? amount.wmul(marginalPrice) : amount.wdiv(marginalPrice);
```

contracts/oracles/yieldspace/YieldSpaceMultiOracle.sol
```diff
@@ -0,0 +1,132 @@
...
+            discountedAmount = (amount * marginalPrice) / 1e18;
+        } else {
+            discountedAmount = (amount * 1e18) / marginalPrice;
```
alcueca, line 129:
```suggestion
            discountedAmount = amount.wdiv(marginalPrice);
```

contracts/oracles/yieldspace/PoolOracle.sol
```diff
@@ -0,0 +1,139 @@
...
+        uint256 minTimeElapsed_
+    ) {
+        require(granularity_ > 1, "GRANULARITY");
+        require((periodSize = windowSize_ / granularity_) * granularity_ == windowSize_, "WINDOW_NOT_EVENLY_DIVISIBLE");
+        windowSize = windowSize_;
+        granularity = granularity_;
+        minTimeElapsed = minTimeElapsed_;
+    }
+
+    // returns the index of the observation corresponding to the given timestamp
```
alcueca, line 56: Please use full natspec for functions, so that docs can be automatically generated.

contracts/oracles/yieldspace/PoolOracle.sol
```diff
@@ -0,0 +1,139 @@
...
+        return ((_getCurrentCumulativeRatio(pool) - oldestObservation.ratioCumulative) * 1e18) / (timeElapsed * 1e27);
+    }
+
+    /// @inheritdoc IPoolOracle
+    function get(address pool) external override returns (uint256 twar) {
+        update(pool);
+        return peek(pool);
+    }
+
+    function _getCurrentCumulativeRatio(address pool) internal view returns (uint256 lastRatio) {
```
alcueca, line 132: Please document also internal functions.

contracts/oracles/yieldspace/YieldSpaceMultiOracle.sol
```diff
@@ -0,0 +1,132 @@
...
+pragma solidity ^0.8.6;
+
+import "./IPoolOracle.sol";
+import "@yield-protocol/yieldspace-interfaces/IPool.sol";
+import "@yield-protocol/vault-interfaces/src/IOracle.sol";
+import "@yield-protocol/utils-v2/contracts/cast/CastBytes32Bytes6.sol";
+import "@yield-protocol/yieldspace-v2/contracts/YieldMath.sol";
+import "@yield-protocol/utils-v2/contracts/access/AccessControl.sol";
+
+contract YieldSpaceMultiOracle is IOracle, AccessControl {
```
alcueca, line 11: Why are YieldSpaceMultiOracle and PoolOracle different contracts?

alcueca (changes requested):
It looks good to me on the whole, although I can't comment on the math (that's for @devtooligan).

Only request is for full natspec throughout, and to make sure the documentation will make this easy to audit.

One point for discussion is whether PoolOracle.sol and YieldSpaceMultiOracle.sol should be merged into a single contract.

contracts/oracles/yieldspace/YieldSpaceMultiOracle.sol
```diff
@@ -0,0 +1,132 @@
...
+        int128 powerValue64 = source.g.mul(source.ts).mul(timeTillMaturity);
+        // scale to 18 dec and convert to regular non-64
+        uint128 powerValue = uint128(powerValue64.mul(wad64x64).toUInt());
+
+        uint256 top = unitPrice.pow(powerValue, uint128(1e18));
+        uint256 bottom = uint128(1e18).pow(powerValue, uint128(1e18)) / 1e18;
+        uint256 marginalPrice = top / bottom;
+
+        if (source.inverse) {
+            discountedAmount = (amount * marginalPrice) / 1e18;
```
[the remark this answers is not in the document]
bbonanno: yeah, leftover from when it was in our codebase (didn't have those funcs)

bbonanno:
> It looks good to me on the whole, although I can't comment on the math (that's for @devtooligan).
> 
> Only request is for full natspec throughout, and to make sure the documentation will make this easy to audit.
> 
> One point for discussion is whether PoolOracle.sol and YieldSpaceMultiOracle.sol should be merged into a single contract.

Yeah, the way I originally saw it is that PoolOracle could also be consumed by 3rd parties, the more consumers the more up-to-date it'll be, and those 3rd parties won't wanna be coupled to the `IOracle` standard...
We could still expose both set of methods though.
I'm on the fence, I think this is better design, but comes at a cost of extra gas due to the extra contract hop.

So yeah, whatever you guys feel it's best, will tackle the rest of the comments in the meantime

alcueca:
The thing is that I always thought of the IOracle interface as consumable by third parties, and anyone can `peek` or `get` any of our oracles.

I see what you mean though, in that calling `poolOracle.get(poolAddress)` and getting a twar is very simple, and maybe some integrators will want the twar for whatever reason, instead of converting from base to fyToken or viceversa.

[I still think that the main use case will be that of converting token amounts from A to B](https://hackernoon.com/getting-prices-right). Maybe merging both contracts and exposing both sets of functions would have the best of both worlds.

alcueca:
I'm happy to approve, whether you merge the contracts or not. I would just like to wait a bit for @devtooligan to check out the math.

contracts/oracles/yieldspace/PoolOracle.sol
```diff
@@ -0,0 +1,146 @@
...
+                }
+            }
+
+            revert MissingHistoricalObservation(pool);
+        }
+    }
+
+    // @inheritdoc IPoolOracle
+    function update(address pool) public override {
+        // populate the array with empty observations (oldest call only)
```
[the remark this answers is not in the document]
devtooligan: alternatively, what if we initialized this in the constructor instead?

contracts/oracles/yieldspace/PoolOracle.sol
```diff
@@ -0,0 +1,146 @@
...
+        for (uint256 i = poolObservations[pool].length; i < granularity; i++) {
+            poolObservations[pool].push();
+        }
+
+        // get the observation for the current period
+        uint256 index = observationIndexOf(block.timestamp);
```
devtooligan, line 101: imagine we're in a 1 hour periodSize.  what happens if update hasn't been called for 3 hours? 🤔

contracts/oracles/yieldspace/PoolOracle.sol
```diff
@@ -0,0 +1,146 @@
...
+    /// @inheritdoc IPoolOracle
+    function get(address pool) external override returns (uint256 twar) {
+        update(pool);
+        return peek(pool);
+    }
+
+    /// @dev computes the most up-to-date TWAR based last calculated ratio and the current cached balances on the pool
+    /// @param pool Address of pool for which the ratio is required 
+    /// @return lastRatio The latest TWAR for `pool`
+    function _getCurrentCumulativeRatio(address pool) internal view returns (uint256 lastRatio) {
```
devtooligan, line 139: can we consider a different name, like `currentRatio` -- to me this sounds too close to `cumulativeRatioLast` which was the initial source of confusion.  i'm so scared of that thing now after the bad experience we had the first time we made this oracle, i even put this warning in the new Pool.sol contract lol:
Screen Shot 2022-05-25 at 3 05 21 PM

devtooligan: btw as you can see we are providing a new thing that you can use `currentCumulativeRatio` in the new pool contract.  there are a couple of breaking changes as well.  and also we are planning to deploy the new contracts in June.  but i want to get through this first set of pr comments on the oracle here before we look at those changes related to the new pool contract.

devtooligan (changes requested):
left comments.  i didnt look at tests yet, i still want to do that.

also, as i alluded to in the comments, there is a new pool.sol contract under review that has some breaking changes with the oracle.  i want to address this pr based on the old(current) Pool.sol first, then we will discuss the new pool ctrct.  It's also not too late to change the new pool ctrct to help out this oracle.  One example is that we are storing `g1` (fees) as a fp4 normal decimal number now (not 64bit) -- but a solution might be that we expose the 64bit g1 for this oracle if it made sense.  But anyways, lets discuss the new pool.sol later.

Screen Shot 2022-05-25 at 3 21 39 PM


As for my vote on whether to combine the two "oracles", it's a tough call.  I did find it confusing.  But I do appreciate separating concerns.  Perhaps whats most confusing about that is that we're trying to fit the TWAR logic into the Oracle interface, maybe we don't need a get and a peek in the TWAR oracle and we use different naming convention for the TWAR?  Once we do that maybe it _does_ make sense to combine them, or perhaps make the TWAR logic a library kinda [like UniV2](https://github.com/Uniswap/v2-periphery/blob/2efa12e0f2d808d9b49737927f0e416fafa5af68/contracts/libraries/UniswapV2OracleLibrary.sol#L16) does.

contracts/oracles/yieldspace/PoolOracle.sol
```diff
@@ -0,0 +1,146 @@
...
+    function observationIndexOf(uint256 timestamp) public view returns (uint256 index) {
+        uint256 epochPeriod = timestamp / periodSize;
+        return epochPeriod % granularity;
+    }
+
+    /// @dev returns the observation from the oldest epoch (at the beginning of the window) relative to the current time
+    /// @param pool Address of pool for which the observation is required
+    /// @return The oldest observation available for `pool`
+    function getOldestObservationInWindow(address pool) public view returns (Observation memory) {
+        if (poolObservations[pool].length == 0) {
```
devtooligan, line 68: if we initalize observations in the constructor we can delete this

contracts/oracles/yieldspace/PoolOracle.sol
```diff
@@ -0,0 +1,146 @@
...
+    /// @inheritdoc IPoolOracle
+    function peek(address pool) public view override returns (uint256 twar) {
+        Observation memory oldestObservation = getOldestObservationInWindow(pool);
+
+        uint256 timeElapsed = block.timestamp - oldestObservation.timestamp;
+        if (timeElapsed > windowSize) {
+            revert MissingHistoricalObservation(pool);
+        }
+
+        if (timeElapsed < minTimeElapsed) {
```
devtooligan, line 122: so this fn will always revert during the first window and also I guess if there was an entire window that passed without any update being called

devtooligan: so i guess if the oracle is in this state the solution is to call `update` then wait the minTimeElapsed before this oracle is useable...right?

contracts/oracles/yieldspace/PoolOracle.sol
```diff
@@ -0,0 +1,146 @@
...
+            uint256 length = poolObservations[pool].length;
+            uint256 oldestObservationIndex;
+            // can't possible overflow
+            for (uint256 i; i < length; ++i) {
+                // no overflow issue. if observationIndex + 1 overflows, result is still zero.
+                oldestObservationIndex = (++observationIndex) % granularity;
+                // If no data exists for this index, check the next one,
+                // this should only happen during the first timeWindow for a given pool
+                // If the elapsedTime is bigger than the timeWindow,
+                // we also check the next one in case only this particular observation is stale
```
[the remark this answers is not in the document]
bbonanno: Yes, it's a circular buffer.
In ideal conditions, you will never have empty/stale slots, so you should always be able to look-up an observation thats about as old as the timeWindow (24hs in the tests).

Problem is, if the data that the looked-up slot has is either empty or it's older than 24hs, the code blows.

Solution (for most cases), is to try the next (23hs) and if that's also bad keep iterating until you find one that works, it's not gonna be 24hs, but should be good enough. There is a last check anyway that whatever observation we use has to be older than a configurable min (5 min in my tests).

If the oracle is widely used and/or we have a bot poking it about once an hour so the slots are always up-to-date, this wouldn't be necessary. I built it cause it's hard to recover if you're missing an observation and the code blows, so with this, I'd only blow for (min window) after someone forces it to take a new observation.

contracts/oracles/yieldspace/PoolOracle.sol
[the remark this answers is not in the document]
bbonanno: Lot's of the comments come from the UniV2 example, so we can write whatever we like, I just kept the same for consistency.
Can't be done in the constructor as we have an array per pool, this is a single oracle contract for multiple pool, one oracle to rule them all :)

contracts/oracles/yieldspace/PoolOracle.sol
```diff
@@ -0,0 +1,146 @@
...
+    /// @return index The index corresponding to the `timestamp`
+    function observationIndexOf(uint256 timestamp) public view returns (uint256 index) {
+        uint256 epochPeriod = timestamp / periodSize;
+        return epochPeriod % granularity;
+    }
+
+    /// @dev returns the observation from the oldest epoch (at the beginning of the window) relative to the current time
+    /// @param pool Address of pool for which the observation is required
+    /// @return The oldest observation available for `pool`
+    function getOldestObservationInWindow(address pool) public view returns (Observation memory) {
```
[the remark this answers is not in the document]
bbonanno: Again, code is adapted from [UniV2](https://github.com/Uniswap/v2-periphery/blob/master/contracts/examples/ExampleSlidingWindowOracle.sol), so I kept it as close as possible to facilitate focusing on the differences for audit purposes, I'd have changed this and many other things, but my reasoning was that keeping it as close as the original would make it easier for other people to understand it

contracts/oracles/yieldspace/PoolOracle.sol
bbonanno: Yes, this code should be in the pool itself imho, or ate the very least in a provided lib like the one you posted in another comment. That way people wouldn't have the same issues we did.

contracts/oracles/yieldspace/PoolOracle.sol
bbonanno: Can't as it's on a per-pool basis

contracts/oracles/yieldspace/PoolOracle.sol
bbonanno: Yes, this is an extra safeguard I added after I did the "walk-thorough" fallback mechanism on the circular buffer, the original code would fail on a missed observation and you'd have to wait at least 1 slot to get it to work again (assuming the next slot was correct). That'd also mean that the first time you use it for a given pool you'd have to wait 24hs for the thing to be usable. My impl only needs 5 min after it's initialised, which in our particular case would be 5min after we do `setSource` on the price oracle (the one that consumes this one)

contracts/oracles/yieldspace/PoolOracle.sol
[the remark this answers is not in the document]
devtooligan: makes sense. would consider cleaning up comment a little

contracts/oracles/yieldspace/PoolOracle.sol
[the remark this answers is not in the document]
devtooligan: ahh yes good point, I was forgetting these are mappings.  can we just add the words "first time" some where in the comment for clarity

contracts/oracles/yieldspace/PoolOracle.sol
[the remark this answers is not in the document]
devtooligan: univ2 didnt get the memo, lol. ok sure lets keep the name

contracts/oracles/yieldspace/PoolOracle.sol
devtooligan: yup

contracts/oracles/yieldspace/PoolOracle.sol
devtooligan: cool

devtooligan (approved):

contracts/foundry-tests/oracles/PoolOracle.t.sol
```diff
@@ -0,0 +1,385 @@
...
+        assertEq(ts, block.timestamp);
+        assertEq(ratio, cumulativeBalancesRatio);
+        assertEq(writes.length, 2);
+    }
+
+    function testPeek() public {
+        address pool = Mocks.mock("IPool");
+
+        // Saturday, 26 February 2022 10:15:28
+        vm.warp(1645870528);
```
devtooligan, line 132: this tests are a bit dense and hard to read.  one thing that could help would be to set some of these constants that you are using more than once to a variable (like you did above with `timestamp`)

contracts/foundry-tests/oracles/PoolOracle.t.sol
```diff
@@ -0,0 +1,385 @@
...
+        assertEq(ts, 1645895728);
+        assertEq(ratio, 6120609608080550173985092961928170);
+    }
+
+    function testGetOldestObservationInWindow() public {
+        address pool = Mocks.mock("IPool");
+
+        // Saturday, 26 February 2022 07:32:48
+        vm.warp(1645860768);
+        IPool(pool).cumulativeBalancesRatio.mock(6081248333687398745122493757731327);
```
devtooligan, line 207: can we set this mocked return value to a variable

contracts/foundry-tests/oracles/PoolOracle.t.sol
```diff
@@ -0,0 +1,385 @@
...
+        assertEq(ts, 0);
+        assertEq(ratio, 0);
+    }
+
+    function testGet() public {
+        address pool = Mocks.mock("IPool");
+
+        // Saturday, 26 February 2022 10:15:28
+        vm.warp(1645870528);
+        IPool(pool).cumulativeBalancesRatio.mock(6081248333687398745122493757731327);
```
devtooligan, line 170: can we set this mocked return value to a variable

contracts/foundry-tests/oracles/PoolOracle.t.sol
```diff
@@ -0,0 +1,385 @@
...
+        IPool(pool).getCache.mock(1102997721918618810252394, 1199964191067551228067732, 1645859234);
+        oracle.update(pool);
+
+        vm.record();
+
+        // Sunday, 27 February 2022 10:15:27 (24h after)
+        vm.warp(1645956927);
+
+        PoolOracle.Observation memory oldestObservation = oracle.getOldestObservationInWindow(pool);
```
devtooligan, line 142: its awesome that we are including the wolframalpha links.  

its still kinda hard to figure out what this test is doing though.  can we add some comments throughout to assist future reviewers.  like you did above in testUpdate

contracts/foundry-tests/oracles/PoolOracle.t.sol
```diff
@@ -0,0 +1,385 @@
...
+
+        // Saturday, 26 February 2022 10:15:28
+        vm.warp(1645870528);
+        IPool(pool).cumulativeBalancesRatio.mock(6081248333687398745122493757731327);
+        IPool(pool).getCache.mock(1102997721918618810252394, 1199964191067551228067732, 1645859234);
+        oracle.update(pool);
+
+        vm.record();
+
+        // Saturday, 26 February 2022 17:15:28 (6h after)
```
devtooligan, line 176: same comment as above about adding some comments to help someone reviewing this for the first time.  like explaining a big about what are all these magic numbers?

alcueca (approved):
:ship: 

Well done!

PR #420, opened 2022-05-24
Addition of authenticated `seize` vault function

iamsahu:
To allow contracts to transfer vaults from one address to another we have added a seize function which moves vault from one user to another without checking whether the owner of the vault is `msg.sender`

### DIFF
--- a/test/092_giver.ts
+++ b/test/092_giver.ts
@@ -86,4 +86,15 @@ describe('Giver', function () {
     const vaultData = await cauldron.vaults(vaultId)
     expect(vaultData['owner']).to.not.eq(owner2)
   })
+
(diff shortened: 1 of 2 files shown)


test/092_giver.ts
```diff
@@ -86,4 +86,15 @@ describe('Giver', function () {
     const vaultData = await cauldron.vaults(vaultId)
     expect(vaultData['owner']).to.not.eq(owner2)
   })
+
+  it('Cannot call the seize function without authentication', async () => {
```
alcueca, line 90:
```suggestion
  it('Cannot seize a vault without authentication', async () => {
```

alcueca (approved):
Just a couple of suggestions on test titles.

PR #423, opened 2022-05-27
HealerModule

Sabnock01:
This PR is for the creation of a `HealerModule` that implements a function `heal` that allows any user to add collateral or repay debt on a specific vault.


contracts/other/backd/HealerModule.sol
```diff
@@ -0,0 +1,76 @@
...
+            fee = ((series.maturity - block.timestamp) * uint256(int256(art)).wmul(borrowingFee)).i128();
+
+        // Update accounting
+        cauldron.pour(vaultId, ink, art + fee);
+
+        // Manage collateral
+        if (ink != 0) {
+            IJoin ilkJoin = getJoin(vault.ilkId);
+            if (ink > 0) ilkJoin.join(vault.owner, uint128(ink));
+            if (ink < 0) ilkJoin.exit(to, uint128(-ink));
```
iamsahu, line 38: Do we need this?

Sabnock01: I'm honestly not sure. The check for negative ink certainly isn't needed though since that's not a possible contingency. 
The logic is from `_pour` though.

Sabnock01: Removed the logic Alberto suggested.

contracts/other/backd/HealerModule.sol
```diff
@@ -0,0 +1,58 @@
+// SPDX-License-Identifier: BUSL-1.1
+pragma solidity 0.8.14;
+
+import "@yield-protocol/vault-interfaces/src/ICauldron.sol";
+import "@yield-protocol/vault-interfaces/src/ILadle.sol";
+import "@yield-protocol/utils-v2/contracts/math/WMul.sol";
+import "@yield-protocol/utils-v2/contracts/cast/CastU256I128.sol";
+import "../../LadleStorage.sol";
+import "../../utils/Giver.sol";
```
alcueca, line 9: Please remove

alcueca (changes requested):
The Cauldron only deals with the accounting, and the Ladle (or its modules) must manage any token operations.

contracts/other/backd/HealerModule.sol
```diff
@@ -0,0 +1,58 @@
...
+        external payable
+    {
+        require (ink >= 0, "Only add collateral");
+        require (art <= 0, "Only repay debt");
+        
+        (bytes12 vaultId, DataTypes.Vault memory vault) = getVault(vaultId_);
+        DataTypes.Series memory series;
+        if (art != 0) series = getSeries(vault.seriesId);
+
+        cauldron.pour(vaultId, ink, art);
```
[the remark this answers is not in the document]
Sabnock01: That makes sense.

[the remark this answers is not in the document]
Sabnock01: You still need to use `vault` in the "Manage collateral" code so you could either keep line 24 or put 
`(, DataTypes.Vault memory vault) = getVault(vaultId_);` as the first line in the block like in the "Manage debt tokens" section.

alcueca (approved):
Great job!

contracts/modules/HealerModule.sol
```diff
@@ -0,0 +1,65 @@
...
+    {
+        require (ink >= 0, "Only add collateral");
+        require (art <= 0, "Only repay debt");
+        (, DataTypes.Vault memory vault) = getVault(vaultId);        
+        cauldron.pour(vaultId, ink, art);
+
+        // Manage collateral
+        if (ink != 0) {
+            IJoin ilkJoin = getJoin(vault.ilkId);
+            ilkJoin.join(vault.owner, uint128(ink));
```
alcueca, line 28: Actually, this is not right.

[Join.join](https://github.com/yieldprotocol/vault-v2/blob/1425bd260c80c0e3dce2d434172c4bc43d9e1ccf/packages/foundry/contracts/Join.sol#L27) will pull assets from the account passed on as a parameter, if it has an allowance to do so.

We don't usually make users `approve` the Join, but if someone did, then anyone could call this module to pull assets from the user into their vault, up to the allowance.

Hardly a disaster, no funds would be lost, still it is an undesiderable behaviour.

A better option would be:
```suggestion
            ilkJoin.join(msg.sender, uint128(ink));
```
In that case, it is responsability of the caller to decide whether to send funds in a previous call, or approve the Join and let it pull the funds.

alcueca (changes requested):
Actually, I found a bug.

alcueca:
I'm noticing only now that there is no natspec, and no description on top of the contract of how this integrates with the Ladle (thinking from the point of view of an auditor here). Would you please fix that?

Sabnock01:
> I'm noticing only now that there is no natspec, and no description on top of the contract of how this integrates with the Ladle (thinking from the point of view of an auditor here). Would you please fix that?

Absolutely.

Sabnock01:
@bbonanno I pushed a fix ([#450](https://github.com/yieldprotocol/vault-v2/pull/450)) for the job which runs `forge snapshot` so that it is compatible with tests for the healer which use mainnet forking yet these still fail because of the `--check` argument because it cannot find previous values for the HealerModule tests. What do you advise?

These tests worked locally for me but I can only assume that maybe had something to do with my individual setup.

alcueca (approved):
Looks god to me now :D

PR #430, opened 2022-06-02
Update CI for use with forge mainnet forking

Sabnock01:
This PR adds the capacity for the CI to use mainnet forking with forge through the use of an RPC url.


alcueca (approved):
So now all of our hardhat tests run on a fork? Neat.

PR #432, opened 2022-06-02
tweak: removed v1 module, and moved deprecated files to `contracts/deprecated`

alcueca:
Removed some crap


alcueca:
The `test05` is failing in `master`. We will fix it or port those tests to foundry.

alcueca:
I'm going to merge, instead of getting you distracted with minutiae

PR #433, opened 2022-06-02
Bring back the `Wand`

iamsahu:


alcueca:
I'm happy with the contract as it is, but I've thought that we should keep these governance helper contracts in the `environments-v2` repo. That would also help you with all this broken CI items ;)

Would you mind closing this PR, and opening a new PR in `environments-v2` with this very contract?

iamsahu:
Closing & moving to the `environments-v2` repo.

PR #434, opened 2022-06-02
Witch V2

bbonanno:


alcueca (approved):
Looks good to me, what would you suggest as further testing before we make this live?

We will put it to audit, but is there some testing suite that you can think of on a mainnet fork?

bbonanno:
> Looks good to me, what would you suggest as further testing before we make this live?
> 
> We will put it to audit, but is there some testing suite that you can think of on a mainnet fork?

yes, I'm doing that as part of the liq bot tests, and it's already paying off, so after I do that my plan is to create a second PR with whatever fixes I need to add

PR #435, opened 2022-06-03
Migrate to yieldspace-tv

alcueca:
Only used for YieldSpaceMultiOracle, which doesn't have unit tests yet even.


alcueca:
@devtooligan, any idea what's breaking here?

devtooligan:
@alcueca Shot in the dark, but could it be because the yieldspace-tv repo is private?  Lemme know if you'd like me to spend more time looking into this.

alcueca:
Ah, that sounds plausible, maybe it's time to make it public.

alcueca:
@devtooligan, at some point we would need to merge this.

alcueca:
Actually, this should be merged only when yieldspace-tv is live, until then we should keep it unmerged.

devtooligan (approved):
Performed a fairly quick visual review.   Mostly reviewed structure and detailed changes.  Did not do side-by-side comparisons of any of the files that were copy-pasted in

lgtm

PR #436, opened 2022-06-03
buildfix: skip failing tests

davidbrai:
these tests are failing for unknown reasons.
skipping for now but alberto says they will be fixed/rewritten later


alcueca (approved):
They will be fixed, I promise. Thank you.

PR #440, opened 2022-06-13
Witch to share dust params with Cauldron

bbonanno:
- Use cauldron params to cater for dust
- small improvements on WitchV2 tests


alcueca (approved):
I see what you mean, we just pick up the `dust` parameter from the Cauldron, instead of replicating it. A bit gassier due to the extra CALL, but I reckon it's warranted for the simpler governance.

bbonanno:
> I see what you mean, we just pick up the `dust` parameter from the Cauldron, instead of replicating it. A bit gassier due to the extra CALL, but I reckon it's warranted for the simpler governance.

yes, but tbh, if we wanted to be super gas efficient we should also replicate the "static" cauldron data like series.

PR #441, opened 2022-06-13
Make artIn param optional on quoter

bbonanno:


alcueca:
What does this block do? It seems wrong to auction a vault from `calcPayout`:
https://github.com/yieldprotocol/vault-v2/blob/f37b10604d2459692a3c2b6ec48691afea06b2f8/packages/foundry/contracts/witch/WitchV2.sol#L369

packages/foundry/contracts/witch/WitchV2.sol
```diff
@@ -375,6 +376,12 @@ contract WitchV2 is AccessControl {
             auction_ = _auction(vault, series, balances, debt);
         }
 
+        // 0 value is to offer a nice API for people that wants to pay all
+        // GT check is to cater for partial buys right before this method executes
+        artIn = (maxArtIn == 0 || maxArtIn > auction_.art)
```
alcueca, line 381: Why not remove the use of zero as a magic number and let the user enter MAX_UINT256?

bbonanno: I think that 0 makes a "nicer" API, but I don't have any other strong justification

bbonanno:
> What does this block do? It seems wrong to auction a vault from `calcPayout`:
> 
> https://github.com/yieldprotocol/vault-v2/blob/f37b10604d2459692a3c2b6ec48691afea06b2f8/packages/foundry/contracts/witch/WitchV2.sol#L369

It's a view only function that's meant to tell bots how much you'd get out of liquidating a vault, even if the vault hasn't been auctioned yet.
It's useful for non auctioned vaults too cause I'm working under the assumption that the collateral will have an initial discount of 5% or something, so you can use this to evaluate 3 things 
- Should I auction + liquidate on the same tx
- Should I just kickstart an auction and wait 
- should I buy a pre-existent auction given the current collateral in offer

alcueca:
Fair enough, I understand now. Let's make it a bit clearer:
 - Include natspec in `_auction`, and maybe change the function name so that it implies that it creates an `Auction` object, but doesn't actually auction anything.
 - Instead of checking `auction_.ink == 0`, check `auction_.start == 0` and put an inline comment explaining that if an auction doesn't exist, we calculate the payout on a hypothetical one.

packages/foundry/contracts/witch/WitchV2.sol
alcueca: I prefer to avoid magic values and edge cases wherever possible. I also think that zero as a magic value is quite dangerous, as you can easily obtain it by accident.

alcueca: I get that this is a view function, but still.

bbonanno: Fair enough, will change it, I don't have any strong reasons other than cosmetics :)

bbonanno: On my defense, the comment about 0 meaning all was there, I just made it work :P

alcueca: I'm happy to agree that I often take decisions that I later regret, so the original comment might have been mine.

bbonanno: LOL, no worries, it's gone now

packages/foundry/contracts/witch/WitchV2.sol
```diff
@@ -354,27 +357,33 @@ contract WitchV2 is AccessControl {
...
+        // If the vault hasn't been auctioned yet, we calculate what values it'd have if it was started right now
+        if (auction_.start == 0) {
             DataTypes.Series memory series = cauldron.series(vault.seriesId);
             DataTypes.Balances memory balances = cauldron.balances(vaultId);
             DataTypes.Debt memory debt = cauldron.debt(
                 series.baseId,
                 vault.ilkId
             );
-            auction_ = _auction(vault, series, balances, debt);
+            auction_ = _calculateAuctionInitialValues(vault, series, balances, debt);
```
alcueca, line 381:
```suggestion
            auction_ = _calcAuction(vault, series, balances, debt);
```

PR #443, opened 2022-06-14
feat/442

bbonanno:
- Add awarenes of other witches
- Add awarenes of non-liquidable pairs to the witch
- Re-use custom error for 'vault already under auction'

### DIFF
--- a/packages/foundry/contracts/witch/WitchV2.sol
+++ b/packages/foundry/contracts/witch/WitchV2.sol
@@ -23,7 +23,7 @@ contract WitchV2 is AccessControl {
     using WDiv for uint256;
     using CastU256U128 for uint256;
 
-    error VaultAlreadyInAuction(bytes12 vaultId, address witch);
+    error VaultAlreadyUnderAuction(bytes12 vaultId, address witch);
     error VaultNotLiquidable(bytes12 vaultId, bytes6 ilkId, bytes6 baseId);
 
     event Auctioned(bytes12 indexed vaultId, uint256 indexed start);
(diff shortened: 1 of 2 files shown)


packages/foundry/contracts/witch/WitchV2.sol
```diff
@@ -110,17 +121,44 @@ contract WitchV2 is AccessControl {
         emit LimitSet(ilkId, baseId, max);
     }
 
+    /// @dev Governance function to set other liquidation contracts that may have taken vaults already.
+    /// @param a The address that may be set/unset as another witch
+    /// @param isWitch Is this address a witch or not
+    function setAnotherWitch(address a, bool isWitch) external auth {
```
alcueca, line 127:
```suggestion
    function setAnotherWitch(address witch, bool isWitch) external auth {
```

packages/foundry/contracts/witch/WitchV2.sol
```diff
@@ -110,17 +121,44 @@ contract WitchV2 is AccessControl {
...
         DataTypes.Vault memory vault = cauldron.vaults(vaultId);
+        if (vault.owner == address(this) || otherWitches[vault.owner]) {
+            revert VaultAlreadyUnderAuction(vaultId, vault.owner);
+        }
         DataTypes.Series memory series = cauldron.series(vault.seriesId);
+        if (ignoredPairs[vault.ilkId][series.baseId]) {
+            revert VaultNotLiquidable(vaultId, vault.ilkId, series.baseId);
+        }
+
+        require(cauldron.level(vaultId) < 0, "Not undercollateralized");
```
alcueca, line 160: Don't you want to make this an error as well?

alcueca (approved):
Minor changes, once fixed feel free to merge.

packages/foundry/contracts/witch/WitchV2.sol
bbonanno: Didn't wanna call it witch cause if isWitch is false makes no sense, I called it a for address cause we don't really know if it's a witch or not, but happy to change it if you feel strong about it

packages/foundry/contracts/witch/WitchV2.sol
bbonanno: I'd change all checks, but I think it should be it's own PR

packages/foundry/contracts/witch/WitchV2.sol
alcueca: Between `a` and `witch` I prefer the latter. You can also use something more generic such as `target` or `value` as long as it's a word.

PR #445, opened 2022-06-16
Remove `pool.buyBasePreview` from `ladle.serve`

### LINKED ISSUE #424: Ladle.serve doesn't need a `buyBasePreview`
`pool.buyBase` returns the fyToken paid, so there is no need for a `buyBasePreview`:

```
    /// @dev Add collateral and borrow from vault, so that a precise amount of base is obtained by the user.
    /// The base is obtained by borrowing fyToken and buying base with it in a pool.
    /// Only before maturity.
    function serve(bytes12 vaultId_, address to, uint128 ink, uint128 base, uint128 max)
        external payable
        returns (uint128 art)
    {
        (bytes12 vaultId, DataTypes.Vault memory vault) = getVault(vaultId_);
        IPool pool = getPool(vault.seriesId);
        
        art = pool.buyBase(to, base, max);
        _pour(vaultId, vault, address(pool), ink.i128(), art.i128());
    }
```

There is a reentrancy risk, mitigated by the fact that only registered vaults can be called.

Sabnock01:
This PR removes `pool.buyBasePreview` from the `ladle.serve` function as `pool.buyBase` returns the FYToken paid. Closes #424

### DIFF
--- a/packages/foundry/contracts/Ladle.sol
+++ b/packages/foundry/contracts/Ladle.sol
@@ -379,9 +379,8 @@ contract Ladle is LadleStorage, AccessControl() {
         (bytes12 vaultId, DataTypes.Vault memory vault) = getVault(vaultId_);
         IPool pool = getPool(vault.seriesId);
         
-        art = pool.buyBasePreview(base);
+        art = pool.buyBase(to, base, max);
         _pour(vaultId, vault, address(pool), ink.i128(), art.i128());
-        pool.buyBase(to, base, max);
     }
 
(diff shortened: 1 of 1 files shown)


alcueca (commented):
Ah, we actually need to run the `preview`:

image

`_pour` actually borrows and amount of fyToken that gets dropped into the pool. Only then we can call `pool.buyBase`. The `*Preview` tells us how much we need to borrow.

alcueca:
Closing as my initial assessment was wrong, and the original implementation is right.

PR #448, opened 2022-06-17
Replace `Test.sol`

### LINKED ISSUE #447: Replace `Test.sol` with `forge-std`
When I originally made [Test.sol}(https://github.com/yieldprotocol/vault-v2/blob/master/packages/foundry/contracts/test/utils/Test.sol) `ds-test` was still installed by default and if memory serves there was some issues with having `forge-std` as a submodule (as opposed to a node module) and it interfering with the TypeScript tests. In which case this should probably be done after all tests have been migrated to Solidity and the TypeScript tests deprecated.

Sabnock01:
This PR replaces `Test.sol` located at `packages/foundry/contracts/test/utils/Test.sol` with the `Test.sol` file provided with `forge-std`. Closes #447


devtooligan:
lgtm

devtooligan (approved):

alcueca (approved):
This was the original reason that led to the refactor into yarn workspaces. I didn't like the hack :D

PR #456, opened 2022-06-30
Rename witch to witchV2 to avoid TS compile issues

bbonanno:


alcueca:
Instead, please remove the old Witch from the repo, and rename WitchV2 as Witch.

PR #457, opened 2022-06-30
Port tests for oracles to Foundry

### LINKED ISSUE #414: Port Oracle tests to foundry
[Here](https://github.com/yieldprotocol/vault-v2/tree/master/test)

image

The other oracle tests can be used as a template.

Sabnock01:
PR for the remaining oracle tests to be ported to Foundry. Closes #414


Sabnock01:
These tests aren't implemented uniformly but I thought I'd leave them as they are presently such that I don't apply a single standard to them only to have to redo them all according to another. So they've been implemented various ways and we can decide which we like best.

packages/foundry/contracts/test/oracles/ChainlinkUSDMultiOracle.t.sol
```diff
@@ -0,0 +1,88 @@
...
+    address public dai = 0x6B175474E89094C44Da98b954EedeAC495271d0F;
+    address public frax = 0x853d955aCEf822Db058eb8505911ED77F175b99e;
+
+    function setUp() public {
+        oracleL1 = new ChainlinkUSDMultiOracle();
+        flagsL2 = new FlagsInterfaceMock();
+        oracleL2 = new ChainlinkL2USDMultiOracle(flagsL2);
+        aggregator = new ChainlinkAggregatorV3MockEx(18);   // for testing non-8-digit source
+
+        vm.createSelectFork('mainnet', 15044600);
```
alcueca, line 30: Nice, I didn't know this was the way to do tests on mainnet forks.

alcueca (approved):
There are a couple of tests missing in `setUp`, though.

packages/foundry/contracts/test/oracles/ChainlinkUSDMultiOracle.t.sol
Sabnock01: Previously, you had to run `forge test` with the `--fork-url` option if you wanted to use a mainnet fork and the `--fork-block-number` option if you didn't want to fork from the latest block. And all of your tests would uniformly be using that block. But now with the new cheatcodes, you can fork on a per-test basis where each one can fork from a different block and you no longer have to pass in those options. The `'mainnet'` is just an alias of the rpc url specified in `foundry.toml` and we're currently assigning through an environment variable.

Issue #453, opened 2022-06-28
Refactor NotionalJoin so that it uses the underlying Join after maturity


alcueca:
After maturity, the first attempt to `exit` at a NotionalJoin exchanges all fCash assets for their underlying, and subsequent calls to `exit` get routed to `_exitUnderlying`.

Instead of storing the underlying in the NotionalJoin, it should be sent to the underlying Join, to reduce asset fragmentation. Afterwards `_exitUnderlying` would just call `underlyingJoin.exit()`.

PR #459, opened 2022-07-02
Use underlying join for fCash joins

calnix:
**Objective:**

After maturity, fCash can be converted to underlying. First call to exit(), and therefore, redeem() will result in the entire fCash holding in NotionalJoin to be converted and sent to the appropriate underlying join (i.e. if fCash converted to DAI, the DAI is sent to the main DAI join in Yield).

Subsequent calls to exit(), will call _exitUnderlying() - which will move the appropriate amount of underlying (e.g. DAI) from the main join to the user. 

### Changes made

**Added `underlyingJoin` as storage variable (line 28)**
- stores address of the deployed underlying join
- for use in IJoin interface

**Functions modified:**
- `constructor`
- `reedem`
- `_exitUnderlying`

`constructor`
- added underlyingJoin on line 41

`reedem`
- line 148 - 152
- transfer underlying to its deployed join by calling `transfer` and `join()` through IJoin interface
- underlyingBalance cast from uint256 to uint128
- storedBalance set to 0 , reflecting zero fCash balance.

`_exitUnderlying`
- line 122 - 124
- fn now calls `join()` on the underlying join to move the tokens to user.

### Testing 
Testing is focused on validating the specific changes made as described above. Other functionality and contracts are assumed to be working as intended. With regard to ERC1155 attributes of maturity and ids, arbitrary values are used as this is not the focus of testing.

We deployed mocked versions of both FCashMock and DAIMock to observe the token flow, from ERC1155 to ERC20. Calls to underlying join contract is mocked via Mocks.sol library functions (as observed in the other foundry tests contracts).


packages/foundry/contracts/other/notional/NotionalJoin.sol
```diff
@@ -181,26 +144,24 @@ contract NotionalJoin is IJoin, ERC1155TokenReceiver, AccessControl() {
         uint256 underlyingBalance = IERC20(underlying).balanceOf(address(this));
         uint256 storedBalance_ = storedBalance;
         accrual = underlyingBalance.wdiv(storedBalance_); // There is a rounding loss here. Some wei will be forever locked in the join.
-        // This becomes now to an exit-only underlying Join.
-        storedBalance = underlyingBalance;
+        
+        // transfer underlying to main join
+        IJoin(underlyingJoin).join(address(this), underlyingBalance.u128());
```
alcueca, line 149: You need to transfer the underlying to the join first.

alcueca (commented):
The approach is sound, there is a `transfer` missing in there.

packages/foundry/contracts/other/notional/NotionalJoin.sol
calnix: Wouldn't that be achieved by calling `join(address(this), underlyingBalance.u128())` ? 
My understanding is that it would move the underlying from the NotionalJoin to itself (underlyingJoin).

calnix: Oh, I understand this now. 

Njoin would `transfer` the underlying to the main join then call `join`.

This is because the join checks the difference btw its own logged balance versus balanceOf (`available`), and moves only the shortfall (`amount - available`) to itself. 

```solidity
        uint256 available = token.balanceOf(address(this)) - _storedBalance; // Fine to panic if this underflows
        unchecked {
            storedBalance = _storedBalance + amount; // Unlikely that a uint128 added to the stored balance will make it overflow
            if (available < amount) token.safeTransferFrom(user, address(this), amount - available);
        }
```

This clicked while I watching your Yield v2 Vault Walkthrough on youtube. Thanks!

calnix:
Tests using Mocks.sol library for underlying join:
[0d53717](https://github.com/yieldprotocol/vault-v2/pull/459/commits/0d53717bd03e39eadaaced9dd4a8e5a50a90238a)

Tests with deployed underlying join:
[1427d10](https://github.com/yieldprotocol/vault-v2/pull/459/commits/1427d10ff045a3eb3f8779ac4382a15bf3fa13f4)

alcueca:
@iamsahu, would you mind doing a review on this PR?

alcueca:
Fixes #453

packages/foundry/contracts/other/notional/NotionalJoin.sol
```diff
@@ -137,32 +116,24 @@ contract NotionalJoin is IJoin, ERC1155TokenReceiver, AccessControl() {
...
 
     /// @dev Transfer `amount` `asset` to `user`
-    function _exit(address user, uint128 amount)
-        internal
-        beforeMaturity
-        returns (uint128)
-    {
+    /// @param user Address of receiver of fCash tokens
+    /// @param amount Amount of ERC1155 tokens
+    function _exit(address user, uint128 amount) internal beforeMaturity returns (uint128) {
```
iamsahu, line 121: Redundant `beforeMaturity` check

packages/foundry/contracts/other/notional/NotionalJoin.sol
```diff
@@ -137,32 +116,24 @@ contract NotionalJoin is IJoin, ERC1155TokenReceiver, AccessControl() {
...
-    function _exitUnderlying(address user, uint128 amount)
-        internal
-        afterMaturity
-        returns (uint128)
-    {
-        storedBalance -= amount;
-        MinimalTransferHelper.safeTransfer(IERC20(underlying), user, amount);
+    /// @param user Recipient of token transfer
+    /// @param amount Amount of underlying tokens to transfer
+    function _exitUnderlying(address user, uint128 amount) internal afterMaturity returns (uint128) {
```
iamsahu, line 130: Redundant `afterMaturity` check

iamsahu (changes requested):
Minor changes. Rest looks good.

packages/foundry/contracts/other/notional/NotionalJoin.sol
calnix: Right, because `exit` has checks if `block.timestamp < maturity`, before executing `_exit`

```solidity
    function exit(address user, uint128 amount) external override auth returns (uint128) {
        if (block.timestamp < maturity) {
            return _exit(user, amount);
        } else {
            if (accrual == 0) redeem(); // Redeem all fCash, switch to underlying join, set accrual.
            return _exitUnderlying(user, uint256(amount).wmul(accrual).u128());
        }
    }
```

calnix: `beforeMaturity`  removed.

packages/foundry/contracts/other/notional/NotionalJoin.sol
calnix: `afterMaturity` check removed.

iamsahu (approved):
LGTM!

alcueca (approved):
Good job on the code changes, and on the foundry tests. I had never seen the syntax for storage changes before.

On the other hand, the code style changes made the review quite difficult, adding a lot of noise to the diff. In the future do the code style changes in a single commit where nothing functional changes, and squash all the actually relevant commits.

That way, the reviewer can see the diff for the code style changes, and verify that nothing actually changes, and then the diff for the functional changes, absent of noise.

The tests I believe are a port of the hardhat tests. It's ok to leave them as is because this code is mostly on mainnet except for this change, but in the future we need to test this on mainnet forks.

alcueca:
Also, please use two or three words for the branch name, people sometimes use `git` on the command line :grimacing:

calnix:
> Good job on the code changes, and on the foundry tests. I had never seen the syntax for storage changes before.
> 
> On the other hand, the code style changes made the review quite difficult, adding a lot of noise to the diff. In the future do the code style changes in a single commit where nothing functional changes, and squash all the actually relevant commits.
> 
> That way, the reviewer can see the diff for the code style changes, and verify that nothing actually changes, and then the diff for the functional changes, absent of noise.
> 
> The tests I believe are a port of the hardhat tests. It's ok to leave them as is because this code is mostly on mainnet except for this change, but in the future we need to test this on mainnet forks.

Apologies for the mess on the diffs due to the commits, I'll take care in the future

iamsahu (approved):

alcueca (approved):
You can remove `023_join_notional.ts` before merging.

PR #462, opened 2022-07-06
Replace witch

bbonanno:
- Had to leave the old witch as all the test setup for the HH tests rely on the Wand, and the Wand relies on the Witch, not sure if there're plans to remove all of that


alcueca (approved):
Looks good to me.

As we move all the unit tests to Foundry, the old fixtures file will be removed, and with that the Wand and old Witch.

PR #463, opened 2022-07-07
Modify `FYToken.sol` to be ERC-5095 compliant

### LINKED ISSUE #451: Implement fyToken.sol as a 5095
[EIP5095](https://github.com/ethereum/EIPs/pull/5095) is still in draft form, but the feeling is that it's current incarnation is not very far from Final. A fyToken implementation would allow us and others in the EIP group to find out the suitability of the standard.

Please use foundry for this.

Sabnock01:
This PR is for the ERC-5095 compliant implementation of `FYToken.sol`. Closes #451


Sabnock01:
This is a very rough draft of the implementation. 

Things to be most likely added/modified: 

- Tests (could include the migration of the FYToken tests to Foundry also in this PR)
- We could make an interface or abstract contract for ERC5095 like the others in `yield-utils-v2`
- Logic in `withdraw` and `redeem` will likely need fixing (The previous logic of `redeem` looks more similar to how `withdraw` is implemented in ERC-5095)

Thoughts on these?

packages/foundry/contracts/FYToken.sol
```diff
@@ -118,15 +140,54 @@ contract FYToken is IFYToken, IERC3156FlashLender, AccessControl, ERC20Permit, C
         accrual_ = accrual_ >= 1e18 ? accrual_ : 1e18; // The accrual can't be below 1 (with 18 decimals)
     }
 
+    ///@dev returns the maximum redeemable amount for the address holder in terms of the principal
+    function maxRedeem(address holder) public view returns (uint256 maxPrincipalAmount) {
+        return _balanceOf[holder];
+    }
+
+    ///@dev returns the amount of underlying redeemable in terms of the principal
+    function previewRedeem(uint256 principalAmount) public view returns (uint256 underlyingAmount) {
```
devtooligan, line 149: this should revert before maturity

packages/foundry/contracts/FYToken.sol
```diff
@@ -118,15 +140,54 @@ contract FYToken is IFYToken, IERC3156FlashLender, AccessControl, ERC20Permit, C
...
+
     /// @dev Burn fyToken after maturity for an amount that increases according to `chi`
     /// If `amount` is 0, the contract will redeem instead the fyToken balance of this contract. Useful for batches.
-    function redeem(address to, uint256 amount) external override afterMaturity returns (uint256 redeemed) {
-        uint256 amount_ = (amount == 0) ? _balanceOf[address(this)] : amount;
-        _burn(msg.sender, amount_);
-        redeemed = amount_.wmul(_accrual());
-        join.exit(to, redeemed.u128());
+    function redeem(uint256 principalAmount, address receiver, address holder) external override afterMaturity returns (uint256 underlyingAmount) {
+        uint256 amount_ = (principalAmount == 0) ? _balanceOf[address(this)] : principalAmount;
```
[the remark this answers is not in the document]
devtooligan: upon further thought, i kinda of like the overload, because if you use type(uint).max (or zero) for something else, then it kinda breaks the whole concept that preview should show exactly what is being redeemed (but at the same time, preview should ignore how much the caller actually holds)

packages/foundry/contracts/FYToken.sol
```diff
@@ -118,15 +140,54 @@ contract FYToken is IFYToken, IERC3156FlashLender, AccessControl, ERC20Permit, C
...
+        emit Redeemed(holder, receiver, amount_, underlyingAmount);
+    }
+
+    ///@dev returns the maximum withdrawable amount for the address holder in terms of the underlying
+    function maxWithdraw(address holder) public view returns (uint256 maxUnderlyingAmount) {
+        return _convertToUnderlying(_balanceOf[holder]);
+    }
+
+    ///@dev returns the amount of the principal withdrawable in terms of the underlying
+    function previewWithdraw(uint256 underlyingAmount) public view returns (uint256 principalAmount) {
```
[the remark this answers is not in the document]
devtooligan: visibility sb external not public

packages/foundry/contracts/FYToken.sol
```diff
@@ -118,15 +140,54 @@ contract FYToken is IFYToken, IERC3156FlashLender, AccessControl, ERC20Permit, C
...
+        return _previewWithdraw(underlyingAmount);
+    }
+
+    ///@dev returns the amount of underlying redeemable in terms of the principal
+    function _previewWithdraw(uint256 underlyingAmount) private view returns (uint256 principalAmount) {
```
devtooligan, line 180: we generally prefer internal to private, more flexible

packages/foundry/contracts/FYToken.sol
devtooligan: visibility sb external when calling an internal fn

packages/foundry/contracts/FYToken.sol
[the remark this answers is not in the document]
devtooligan: it sb a overloaded external fn that looks up the callers bal and then passes it in to the users ...

if we go this route, then redeem / withdraw sb external fns that call internal fns -- this is different than the implementation we did but i think thats right https://github.com/yieldprotocol/ERC5095/pull/1/files

packages/foundry/contracts/FYToken.sol
```diff
@@ -118,15 +140,54 @@ contract FYToken is IFYToken, IERC3156FlashLender, AccessControl, ERC20Permit, C
...
+        return _convertToPrincipal(underlyingAmount);
```
devtooligan, line 181: this sb exactly the same fn as redeem, but instead of "they give us a certain principal to redeem x underlying" its "they give us x principal to redeem a certain amount of underlying" so this should have the same calcs w accrual as redeem so:

```solidity
     principalAmount = underlyingAmount.wdiv(_accrual());
```

packages/foundry/contracts/FYToken.sol
```diff
@@ -86,6 +86,28 @@ contract FYToken is IFYToken, IERC3156FlashLender, AccessControl, ERC20Permit, C
         emit FlashFeeFactorSet(flashFeeFactor_);
     }
 
+    ///@dev Converts the amount of the principal to the underlying
```
devtooligan, line 89: note for above: @alcueca mentioned we want to start moving the interfaces out of that `vault-interfaces` repo and into the repo of the contract.  this might be a time to start that and move the IFYToken file to this repo

devtooligan:
@Sabnock01 great first draft!

please take a look at our implementation and try to model it on that: https://github.com/yieldprotocol/ERC5095/pull/1/files  esp read some of the comments about reverting etc.  u can use the modifier if u want, or inline it, i dont have a strong preference personally.

although, in that version we made redeem/withdraw public fns and the previews were external+internal....but looking at this one, I almost feel like it should be the opposite.  or else both external+internal....  lets get everything else working and we can make the call on teh public vs external+internal later

_needs tests_ as you noted.  lets get some unit tests in on the next pass

fyi -- i didn't dig deeper or review carefully the `amount_.wmul(_accrual())` math -- i just assumed you ported it over correctly

> We could make an interface or abstract contract for ERC-5095 like the others in yield-utils-v2
@Sabnock01 what do u mean by this? i'm interested

devtooligan:
> Moreover logic between IFYToken.sol's redeem and the ERC-5095's redeem needs addressing such that the function signatures do not conflict. I'm thinking we could just change the signature in IFYToken.sol to be the same as ERC-5095?

If we go with my proposed solution (overload `redeem()`) then we don't need to change this AND we stay backwards compatible...(we would need to add a second `redeem()` to the interface though)

Sabnock01:
> > We could make an interface or abstract contract for ERC-5095 like the others in yield-utils-v2
> 
> 
> @Sabnock01 what do u mean by this? i'm interested

Pretty much the exact file of yours you linked to of ERC5095. Similar to ERC20Permit, inherited by `FYToken.sol`. 
But wouldn't that be better fit in `yield-utils-v2` similar to `ERC20.sol` or `ERC20Permit.sol`?

packages/foundry/contracts/FYToken.sol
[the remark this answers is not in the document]
alcueca: So, this is one of my dirty hacks, to allow the fyToken to redeem what was sent to it, without necessarily having to know how much is it.

In a batch, when we burn some pool tokens, we might want to redeem the fyToken part if it is mature, but it will be impossible to know exactly how much is that.

I think that a second `function redeem(address receiver) external afterMaturity returns (uint256 underlyingAmount)` is indeed much cleaner. The Ladle doesn't ever call `redeem` itself (please check!), so we could get away with refactoring it and letting the frontend guys know.

packages/foundry/contracts/FYToken.sol
```diff
@@ -118,15 +140,54 @@ contract FYToken is IFYToken, IERC3156FlashLender, AccessControl, ERC20Permit, C
...
+    }
+
+    ///@dev returns the amount of underlying redeemable in terms of the principal
+    function previewRedeem(uint256 principalAmount) public view returns (uint256 underlyingAmount) {
+        return _previewRedeem(principalAmount);
+    }
+
+    ///@dev returns the amount of underlying redeemable in terms of the principal
+    function _previewRedeem(uint256 principalAmount) private view returns (uint256 underlyingAmount) {
+        return _convertToUnderlying(principalAmount);
```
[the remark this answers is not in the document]
Sabnock01: `return _convertToUnderlying(principalAmount.wmul(_accrual()));`?

packages/foundry/contracts/FYToken.sol
Sabnock01: Created.

Sabnock01:
There is still some logic to implement. (I noticed I hadn't implemented tests for the `IERC5095.sol` logic yet and keep getting `"EvmError: Revert"` when I try to call those functions in the tests; will continue looking into it) Nonetheless, I figure this is probably a good spot to give it another review since all of the former testing logic has now been converted. 

But other than that the only things left should be: 
1. Correct `redeem`(new logic) if need be
2. Write the tests for the new logic

packages/foundry/contracts/FYToken.sol
```diff
@@ -86,6 +86,36 @@ contract FYToken is IFYToken, IERC3156FlashLender, AccessControl, ERC20Permit, C
...
+    }
+
+    ///@dev Converts the amount of the principal to the underlying
+    ///Before maturity, returns amount as if at maturity.
+    function _convertToUnderlying(uint256 principalAmount) internal returns (uint256 underlyingAmount) {
+        if (chiAtMaturity == CHI_NOT_SET) {
+            return principalAmount;
+        } else {
+            return _accrual();
```
alcueca, line 100: Wouldn't this be `underlyingAmount.wmul(_accrual())`?

Does the EIP say anything about rounding?

Could you please code the unit tests that catch this bug first, and then fix it?

packages/foundry/contracts/FYToken.sol
```diff
@@ -118,6 +148,38 @@ contract FYToken is IFYToken, IERC3156FlashLender, AccessControl, ERC20Permit, C
...
+    }
+
+    ///@dev returns the amount of underlying redeemable in terms of the principal
+    function previewRedeem(uint256 principalAmount) external override beforeMaturity returns (uint256 underlyingAmount) {
+        return _previewRedeem(principalAmount);
+    }
+
+    ///@dev returns the amount of underlying redeemable in terms of the principal
+    function _previewRedeem(uint256 principalAmount) internal returns (uint256 underlyingAmount) {
+        underlyingAmount = principalAmount.wmul(_accrual());
```
alcueca, line 163: We probably want to use `_convertToUnderlying` here to avoid duplicating logic.

packages/foundry/contracts/FYToken.sol
```diff
@@ -118,6 +148,38 @@ contract FYToken is IFYToken, IERC3156FlashLender, AccessControl, ERC20Permit, C
...
+    function redeem(uint256 principalAmount, address receiver, address holder) external override afterMaturity returns (uint256 underlyingAmount) {
+        if (msg.sender != holder) {
+            uint256 allowed = _allowance[holder][msg.sender];
+
+            if (allowed != type(uint256).max) _allowance[holder][msg.sender] = allowed - principalAmount;
+        }
+
+        uint256 amount_ = (principalAmount == 0) ? _balanceOf[address(this)] : principalAmount;
+        _burn(holder, amount_);
+        underlyingAmount = amount_.wmul(_accrual());
```
alcueca, line 177: We probably want to use `_convertToUnderlying` here to avoid duplicating logic.

packages/foundry/contracts/FYToken.sol
```diff
@@ -129,6 +191,35 @@ contract FYToken is IFYToken, IERC3156FlashLender, AccessControl, ERC20Permit, C
...
     }
 
+    ///@dev returns the maximum withdrawable amount for the address holder in terms of the underlying
+    function maxWithdraw(address holder) external override returns (uint256 maxUnderlyingAmount) {
+        return _convertToUnderlying(_balanceOf[holder]);
+    }
+
+    ///@dev returns the amount of the principal withdrawable in terms of the underlying
+    function previewWithdraw(uint256 underlyingAmount) external override beforeMaturity returns (uint256 principalAmount) {
+        return _previewWithdraw(underlyingAmount);
```
alcueca, line 201: You can remove `_previewWithdraw` and use `_convertToPrincipal` here.

packages/foundry/contracts/FYToken.sol
```diff
@@ -129,6 +191,35 @@ contract FYToken is IFYToken, IERC3156FlashLender, AccessControl, ERC20Permit, C
...
+    ///@dev returns the amount of underlying redeemable in terms of the principal
+    function _previewWithdraw(uint256 underlyingAmount) internal returns (uint256 principalAmount) {
+        principalAmount = underlyingAmount.wdiv(_accrual());
+    }
+
+    function withdraw(uint256 underlyingAmount, address receiver, address holder) external override afterMaturity returns (uint256 principalAmount) {
+        if (msg.sender != holder) {
+            uint256 allowed = _allowance[holder][msg.sender];
+
+            if (allowed != type(uint256).max) _allowance[holder][msg.sender] = allowed - principalAmount;
```
alcueca, line 213: `principalAmount` will always be 0 here.

if `underlyingAmount` is 0, then we skip the allowance check and `principalAmount = _balanceOf[address(this)]`
else, we `principalAmount = _convertToPrincipal(underlyingAmount)` and then check for allowance of `principalAmount`

Could you please code the unit tests that catch this bug first, and then fix it?

alcueca (changes requested):
Caught a couple of bugs, and some things can be simplified a bit.

packages/foundry/contracts/FYToken.sol
Sabnock01: It would, yes.

packages/foundry/contracts/FYToken.sol
Sabnock01: I think I found another problem. If the allowance is subtracted before the call to burn, and say hypothetically the amount provided is the same as the allowance, then burn will revert on insufficient allowance since it checks it there.

Sabnock01: Since `_burn` checks the allowance and resets it internally, I think we can avoid this check.

alcueca: Good point

alcueca:
The logic looks solid to me, a few style fixes now, please:
We are mixing the naming conventions from the old fyToken.sol and EIP5095, let's move to the 5095 naming style completely:
`to` -> `receiver`
`from` -> `holder`
`amount` -> `principalAmount` or `underlyingAmount`
any others?

Also a few functions miss natspec, please add it.

packages/foundry/contracts/FYToken.sol
```diff
@@ -86,6 +86,36 @@ contract FYToken is IFYToken, IERC3156FlashLender, AccessControl, ERC20Permit, C
...
+        return _convertToPrincipal(underlyingAmount);
+    }
+
+    ///@dev Converts the amount of the underlying to the principal
+    ///Before maturity, returns amount as if at maturity.
+    function _convertToPrincipal(uint256 underlyingAmount) internal returns (uint256 princpalAmount) {
+        if (chiAtMaturity == CHI_NOT_SET) {
+            return underlyingAmount;
+        } else {
+            return underlyingAmount.wdiv(_accrual());
```
alcueca, line 115: `wmul` (used in `_convertToUnderlying`) and `wdiv` (used in `_convertToPrincipal`) both round down, meaning that `x != _convertToUnderlying(_convertToPrincipal(x))` and `x != _convertToPrincipal(_convertToUnderlying(x))`. Such double reversions happen in some integrations and could be troublesome.

In this case, we are fortunate because it is clear that if we round down on `_convertToUnderlying` and up on `_convertToPrincipal` we will always be rounding in favor of the contract, and the double reversions will match.

You can find `wdivup` in `utils-v2`.

alcueca (changes requested):
Just one change, which should be easy by now. Then we are done.

alcueca (approved):
Let's merge, and I'll announce this on the 5095 Discord (and maybe twitter) tonight.

PR #466, opened 2022-07-12
feat: skip proportion calc for infinite duration

alcueca:
Entering `type(uint32).max` as the `duration` for any `line` is now interpreted as infinite duration, and the proportion calculation is skipped.

The result is that if you want to configure the Witch to skip auctions and liquidate positions in a Compound style you do the following:
`line.proportion = 0.25`
`line.initialOffer = 0.8`
`line.duration = type(uint32).max`

Now imagine the scenario where there is a vault collateralized with 3 ETH at 1000 ETH/DAI, a debt of 2000 DAI, and a collateralization ratio of 150%. Right on the edge.

Price drops to 990 ETH/DAI, vault is undercollateralized, liquidations kick in.

0.75 ETH collateral is auctioned for 500 DAI debt. Only that the initial offer is 0.8 and the duration is infinite, so 0.6 ETH is auctioned for 500 DAI debt. The liquidation is profitable (collateral worth 594 DAI for repaying 500 DAI debt) and is carried through.

The vault remains as 2.4 ETH collateral and 1500 DAI debt at 950 ETH/DAI. Collateralization level is 2.4 * 990 / 1500 = 1.584%


alcueca:
Merging so that we can upload it.

PR #469, opened 2022-07-15
Feat/deprecate vault interfaces

alcueca:
Deprecate the vault-interfaces repo and bring the interfaces into vault-v2


.gitmodules
```diff
@@ -1,21 +1,15 @@
...
 	url = https://github.com/makerdao/dss-interfaces
-[submodule "packages/foundry/lib/yieldspace-tv"]
-	path = packages/foundry/lib/yieldspace-tv
-	url = https://github.com/yieldprotocol/yieldspace-tv
 [submodule "packages/foundry/lib/forge-std"]
 	path = packages/foundry/lib/forge-std
 	url = https://github.com/foundry-rs/forge-std
+[submodule "packages/foundry/lib/yieldspace-tv"]
+	path = packages/foundry/lib/yieldspace-tv
+	url = https://github.com/yieldprotocol/yieldspace-tv
```
bbonanno, line 15: So git resolves the circular dependency?

bbonanno:
this also adds a lot of Uniswap contracts, ignoring scope of the PR, wouldn't it be better to use a foundry dependency when possible?

devtooligan (approved):
lgtm after addressing @bbonanno's questions

.gitmodules
alcueca: The circular dependencies are resolved with submodules, npm and remappings. I'm not very clear on how I managed to do it now.

alcueca: Ok, so looking back into it, for foundry to work I do forge install and a remapping.

For hardhat, I do npm install.

PR #478, opened 2022-08-09
Npm bumpski

devtooligan:


alcueca:
Please use `rc` versions to publish packages that don't pass CI.

I published with `0.18.2-rc.0`, next one will be `0.18.2-rc.1`, and when we are certain the package works, then it gets updated to `0.18.2`

PR #488, opened 2022-09-21
Delegate YieldSpaceMultiOracle math to PoolOracle

bbonanno:


packages/foundry/contracts/oracles/yieldspace/YieldSpaceMultiOracle.sol
```diff
@@ -77,20 +60,13 @@ contract YieldSpaceMultiOracle is IOracle, AccessControl {
         uint256 amount
     ) external view override returns (uint256 value, uint256 updateTime) {
         //solhint-disable-next-line not-rely-on-time
-        updateTime = block.timestamp;
-
-        if (base == quote) return (amount, updateTime);
+        if (base == quote) return (amount, block.timestamp);
```
alcueca, line 63: This is not really true, is it? `peek` doesn't update anything, so I guess `updateTime` is somewhere in the past. This is probably one of the few oracles in which this return value is useful.

alcueca: ChainlinkMultiOracle [relays the updateTime](https://github.com/yieldprotocol/vault-v2/blob/821c324928a91bb6451dfd03692d6d5377a7e1f1/packages/foundry/contracts/oracles/chainlink/ChainlinkMultiOracle.sol#L80) from the chainlink aggregator, doesn't it?

Isn't the YieldSpaceMultiOracle returning the conversion from some point in the past when you call `peek`?

PR #490, opened 2022-09-23
Witch bugfix

bbonanno:
Testing the bot with all the main net history we discovered that some vaults made the auction calculation fail.
Those vaults are broken in many other ways, but made me realised that any vault for which the debt is below the min debt would make the auction calculation fail.
I think it needs to be addressed cause otherwise you can never raise the min debt unless you ensure there are 0 vaults below the new min, which I think is probably not an option.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,6 +3,7 @@
 *~
 *.swp
 *.swo
+.idea
 
 # Packaging
 *.tar.gz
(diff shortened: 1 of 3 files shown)


devtooligan (commented):
lgtm but i would like @alcueca to take a look before we approve/merge

alcueca (approved):
The diff really butchered this change, you only added a conditional like:
```
if (balances.art > min)  { do as before }
else { auction everything }
```
Right?

bbonanno:
> The diff really butchered this change, you only added a conditional like:
> 
> ```
> if (balances.art > min)  { do as before }
> else { auction everything }
> ```
> 
> Right?

Correct

PR #494, opened 2022-10-04
feat: allow to clear auctions

alcueca:
In the case that there are two Witches, and they haven't been correctly set to not steal vaults from each other, we would arrive to a state where in a Witch or both remain auctions that can't operate in the vaults that they are auctioning, and can't be cleared either.

This can happen as well when migrating from Witch v1 to v2, or in other future cases where we allow to `seize` vaults.


bbonanno (approved):
Good one, edge case but worth having the recourse just in case

marcomariscal (approved):

PR #496, opened 2022-10-21
CrabStrategy oracle

iamsahu:


packages/foundry/contracts/oracles/crab/CrabOracle.sol
```diff
@@ -0,0 +1,142 @@
...
+
+/**
+ * @title CrabOracle
+ * @notice Oracle to fetch Crab-ETH exchange amounts
+ */
+contract CrabOracle is IOracle, AccessControl {
+    using CastBytes32Bytes6 for bytes32;
+    ICrabStrategy crabStrategy;
+    IOracle uniswapV3Oracle;
+    bytes6 public constant weth = 0x303000000000;
```
Sabnock01, line 38: Do we normally hard code the ilkId like this?

packages/foundry/contracts/oracles/crab/CrabOracle.sol
```diff
@@ -0,0 +1,142 @@
...
+     * @notice Retrieve the value of the amount at the latest oracle price.
+     * Only `crabId` and `ethId` are accepted as asset identifiers.
+     */
+    function peek(
+        bytes32 base,
+        bytes32 quote,
+        uint256 baseAmount
+    )
+        external
+        view
```
Sabnock01, line 75: Simply adding `view` is the only differentiation between `get` and `peek`?

packages/foundry/contracts/test/oracles/CrabOracleTest.t.sol
```diff
@@ -0,0 +1,60 @@
...
+import "forge-std/src/Test.sol";
+import "@yield-protocol/utils-v2/contracts/access/AccessControl.sol";
+import "../../oracles/crab/CrabOracle.sol";
+import {ICrabStrategy} from "../../oracles/crab/CrabOracle.sol";
+import "../../oracles/uniswap/UniswapV3Oracle.sol";
+
+contract CrabOracleTest is Test {
+    CrabOracle public crabOracle;
+    bytes6 baseId = 0x323900000000;
+    bytes6 quoteId = 0x303000000000;
```
Sabnock01, line 13: You can import these from `TestConstants.sol`

packages/foundry/contracts/test/oracles/CrabOracleTest.t.sol
```diff
@@ -0,0 +1,60 @@
...
+    UniswapV3Oracle uniswapV3Oracle =
+        UniswapV3Oracle(0x358538ea4F52Ac15C551f88C701696f6d9b38F3C);
+
+    function setUp() public {
+        crabOracle = new CrabOracle();
```
Sabnock01, line 18: Please follow the pattern in the other tests of using `vm.createSelectFork('mainnet', <block_number>)` for the first line in `setUp`.

This way it plays nice with the CI.

packages/foundry/contracts/test/oracles/CrabOracleTest.t.sol
```diff
@@ -0,0 +1,60 @@
...
+            0x313900000000,
+            ICrabStrategy(0x3B960E47784150F5a63777201ee2B15253D713e8),
+            IOracle(address(uniswapV3Oracle))
+        );
+
+        vm.startPrank(0x3b870db67a45611CF4723d44487EAF398fAc51E3);
+        uniswapV3Oracle.grantRole(
+            0xe4650418,
+            0x3b870db67a45611CF4723d44487EAF398fAc51E3
+        );
```
Sabnock01, line 31: This is fine though I usually prefer making a state variable of the address at the top so someone reading the test knows exactly what it represents. So in this case something like `address timelock = <that_address>`. I also call the function selector for the same reason instead of hard coding it.

packages/foundry/contracts/test/oracles/CrabOracleTest.t.sol
```diff
@@ -0,0 +1,60 @@
...
+    }
+
+    function testPeek() public {
+        (uint256 amount, ) = crabOracle.peek(
+            bytes32(baseId),
+            bytes32(quoteId),
+            1e18
+        );
+        emit log_named_uint("Crab in ETH Value", amount);
+        // assertEq(amount, 1000000000000000000);
```
Sabnock01, line 48: Why are these assert statements commented out? (If you pin the tests to a specific block how I suggested above, you can leave them uncommented and they'll work indefinitely)

Sabnock01 (changes requested):
Pretty good! Some minor suggestions. Also, do you think you should test `get` as well?

packages/foundry/contracts/oracles/crab/CrabOracle.sol
iamsahu: In `IOracle` get didn't have `view` as in some of the oracles there is a storage change expected.

packages/foundry/contracts/oracles/crab/CrabOracle.sol
iamsahu: We usually don't however the weth is unlikely to change so have gone ahead with the hardcoded value

Sabnock01 (commented):
Looks good to me. Add those base id's to `TestConstants.sol` and call the selectors on the functions please and I'll approve!

Sabnock01 (approved):
Beautiful!

packages/foundry/contracts/oracles/crab/CrabOracle.sol
```diff
@@ -0,0 +1,151 @@
...
+        bytes6 oSQTH_,
+        bytes6 ethId_,
+        ICrabStrategy indexed crabStrategy_,
+        IOracle indexed uniswapV3Oracle_
+    );
+
+    /**
+     * @notice Set crabstrategy & uniswap source
+     */
+    function setSource(
```
alcueca, line 53: You can skip this function and set the source on the constructor, which also means that you don't need to inherit AccessControl anymore. Doing that you can also set all state variables to `immutable`.

packages/foundry/contracts/oracles/crab/CrabOracle.sol
```diff
@@ -0,0 +1,151 @@
...
+            .getVaultDetails();
+        // Get oSQTH price from uniswapOracle
+        (uint256 oSQTHPrice, uint256 lastUpdateTime) = uniswapV3Oracle.peek(
+            oSQTHId,
+            ethId,
+            1e18
+        );
+        require(lastUpdateTime != 0, "Incomplete round");
+        // Crab Price calculation
+        crabPrice =
```
alcueca, line 147: A explanation on this formula would be nice.

packages/foundry/contracts/oracles/crab/CrabOracle.sol
```diff
@@ -0,0 +1,151 @@
...
+        require(
+            (base == crabId && quote == ethId) ||
+                (base == ethId && quote == crabId),
+            "Source not found"
+        );
+
+        if (base == crabId) {
+            quoteAmount = (_getCrabPrice() * baseAmount) / 1e18;
+        } else if (quote == crabId) {
+            quoteAmount = (baseAmount * 1e18) / _getCrabPrice();
```
alcueca, line 127: Two comments explaining what the `1e18` represents would be nice.

alcueca (changes requested):

alcueca (approved):
LGTM

alcueca:
We will fix the NotionalJoin tests now with the ETH fix.

PR #500, opened 2022-10-31
Reuse code in `redeem`

alcueca:
There was an instance where the accrual calculation was done manually, instead of calling `_convertToUnderlying`


Sabnock01 (approved):
Good catch!

PR #506, opened 2022-11-22
fix: receive and wrap ETH from Notional

alcueca:
Draft fix for Notional sending unwrapped ETH on redemption to the NotionalJoin.

Tests still not working because Notional is hard.

Published as 0.16.7-rc.0

PR #507, opened 2022-11-25
Join Test Harness

alcueca:
This is a test harness for plain Join contracts, which can be used as a template for other test harnesses.

The `Join.t.sol` test file reads three environment variables:
NETWORK: Which can be `LOCALHOST`, `MAINNET` or `ARBITRUM`. This drives whether to fork or not, and what to fork.
MOCK: Which can be `true` or `false`. If `true`, the test file will deploy a new Join for testing. If `false`, the test file will read the `JOIN` environment variable to know which live Join to test.
JOIN: If `MOCK == false`, this is the Join that will be tested.

Additionally, there is a `joinHarness.sh` bash script that calls `forge test` for each one of the Joins in an array.

The only downside is that you can't just call `forge test` anymore. To get the previous behaviour you would need to call `MOCK=true NETWORK=LOCALHOST forge test`.


alcueca:
The foundry tests that are failing are the NotionalJoin ones, that I'll fix in [this branch](https://github.com/yieldprotocol/vault-v2/pull/506).

Sabnock01 (approved):
Awesome PR!!! I think we wait until #506 is merged and the Foundry cheatcode is added (which should be shortly) before we merge this one.

PR #510, opened 2022-12-13
fix: ILadle.sol import error

Sabnock01:
I don't know why this works but when I made these changes in `strategy-v2`, the tests all passed.

### DIFF
--- a/README.md
+++ b/README.md
@@ -80,7 +80,7 @@ $ npm run hardhat:test
 ```
 
 ## Bug Bounty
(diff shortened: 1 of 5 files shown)


alcueca:
If we need to remove the hardhat tests I'm all up for it.

Sabnock01:
> If we need to remove the hardhat tests I'm all up for it.

They will be removed one way or another eventually.

alcueca:
There must be something in the `remappings.txt` in this repo messing things up.

alcueca:
Removing the `dss` import from ILadle, and `forwardDaiPermit` with it also fixes my problems in `strategy-v2`, without messing up with things here that might blow up in places like `environments-v2`.

alcueca (approved):

PR #512, opened 2022-12-21
Oracle harness

Sabnock01:


alcueca (approved):
That was a lot of work, thanks for doing this.

PR #517, opened 2023-01-23
ZenBull strategy oracle

iamsahu:


packages/foundry/contracts/oracles/opyn/ZenBullOracle.sol
```diff
@@ -0,0 +1,176 @@
...
+        uint256 bullCrabBalance = zenBullStrategy.getCrabBalance();
+        (uint256 crabEthBalance, uint256 craboSqthBalance) = zenBullStrategy
+            .getCrabVaultDetails();
+        uint256 crabTotalSupply = crabStrategy.totalSupply();
+        uint256 bullTotalSupply = zenBullStrategy.totalSupply();
+        (, int24 tick, , , , , ) = osqthWethPool.slot0();
+        int256 osqthWethPrice = wadDiv(
+            1e18,
+            wadPow(10001e14, int256(tick) * 1e18)
+        );
```
alcueca, line 153: For clarity, I prefer to use `using` and name some constants:
```
using wadDiv for int256
using wadPow for int256
int256 constant public ONE = 1e18;
int256 constant public BASE = 10001e14;
...
(, int24 tick_, , , , , ) = osqthWethPool.slot0();
int256 tick = int256(tick_) * 1e18 // Normalize tick
int256 osqthWethPrice = ONE.wadDiv(BASE.wadPow(tick))
```

packages/foundry/contracts/oracles/opyn/ZenBullOracle.sol
```diff
@@ -0,0 +1,176 @@
...
+        (, tick, , , , , ) = wethUsdcPool.slot0();
+        int256 wethUsdcPrice = wadDiv(
+            1e18,
+            wadPow(10001e14, int256(tick) * 1e18)
+        );
+
+        int256 crabUsdcValue = wadMul(int256(crabEthBalance), wethUsdcPrice) -
```
alcueca, line 160: As you make the math clearer, please use `.i256()` for casting.

packages/foundry/contracts/oracles/opyn/ZenBullOracle.sol
```diff
@@ -0,0 +1,176 @@
...
+        uint256 baseAmount
+    ) private view returns (uint256 quoteAmount, uint256 updateTime) {
+        if (base == zenBullId && quote == usdcId) {
+            quoteAmount = (_getZenBullPrice() * baseAmount) / 1e18;
+        } else if (base == usdcId && quote == zenBullId) {
+            quoteAmount = (baseAmount * 1e18) / _getZenBullPrice();
+        } else {
+            revert ZenBullOracleUnsupportedAsset();
+        }
+        updateTime = block.timestamp;
```
alcueca, line 131: This is not true, is it? The updateTime would be the time of the last observation, if we can get that.

packages/foundry/contracts/test/oracles/ZenBullOracleTest.t.sol
```diff
@@ -0,0 +1,59 @@
+// SPDX-License-Identifier: BUSL-1.1
```
alcueca, line 1: Please deploy on a fork, and use that to code a test harness that verifies that the ZenBullPrice obtained is within an order of magnitude of something. If ZenBull is expected to stay within 0.1 ETH and 10 ETH most of the time, that would be a good range to test against.

alcueca (changes requested):
The math can be made clearer, and casting fixed. A test harness is also a good idea at this point.

PR #518, opened 2023-01-24
USDT Join

Sabnock01:


Sabnock01:
Actually if the join already has a discrepancy between it's total supply and internal counter, a join pull or join combination will fail.

alcueca:
I reduced the code a bit by inheriting from FlashJoin and extracting the fee reversal formula into its own function, I hope you don't mind. Reusing the FlashJoin code means that there is much less code to audit.

We could further generalize this by having a FeeJoin.sol, which has a zero fee implementation of `_reverseFee`, and then ThetherJoin inherits from FeeJoin and implements `_reverseFee` according to the USDT needs, but it seems to me that that would be a premature generalization.

I don't get the comment on a discrepancy making the `join` fail. Note that `storedBalance` can never be higher than `asset.balanceOf(address(join))`. Am I missing anything?

Sabnock01:
> I reduced the code a bit by inheriting from FlashJoin and extracting the fee reversal formula into its own function, I hope you don't mind. Reusing the FlashJoin code means that there is much less code to audit.
> 
> We could further generalize this by having a FeeJoin.sol, which has a zero fee implementation of `_reverseFee`, and then ThetherJoin inherits from FeeJoin and implements `_reverseFee` according to the USDT needs, but it seems to me that that would be a premature generalization.
> 
> I don't get the comment on a discrepancy making the `join` fail. Note that `storedBalance` can never be higher than `asset.balanceOf(address(join))`. Am I missing anything?

The refactor is fine by me. 

As for the test case, I tried transferring an initial amount to the join in the `WithFees.setUp()` and it failed the join push and join combination.

packages/foundry/contracts/other/tether/TetherJoin.sol
```diff
@@ -0,0 +1,37 @@
...
+import "@yield-protocol/utils-v2/contracts/token/TransferHelper.sol";
+
+contract TetherJoin is FlashJoin {
+    using TransferHelper for IERC20;
+
+    constructor(address asset_) FlashJoin(asset_) {}
+
+    /// @dev Calculate the amount of `asset` that needs to be sent to receive `amount` of USDT.
+    function _reverseFee(uint256 amount) internal view returns (uint256) {
+        return amount * 1000000 / (1000000 - (IUSDT(asset).basisPointsRate() * 100));
```
alcueca, line 17: I'm not sure of the math here.

The original code is:
`received = sent.sub(_value.mul(basisPointsRate).div(10000))`

So the reverse is:
```
received = sent - sent.mul(basisPointsRate).div(10000)
received = sent - sent *↓ (basisPointsRate /↓ 10000)
received = sent *↓ (1 - (basisPointsRate /↓ 10000))
received = sent *↓ (10000 - basisPointsRate)
sent = received /↑ (10000 - basisPointsRate)
```
Note that when we reverse a calculation, taking rounding into account, the regular *↓ becomes a /↑. The div(10000) is also a /↓ so I think it doesn't change the rounding.

I think that in your implementation the 1000000 and 100 factor changes are a bit confusing. To make implementation easiest I would import WDivUp from yield-utils and convert into a fixed point factor with 18 decimals.

`sent = received.wdivup(1e18 - basisPointsRate * 1e14)`

alcueca: Also, take into account that the fee is capped at `IUSDT(asset).maximumFee()`, so the above is:

`sent = min(received.wdivup(1e18 - basisPointsRate * 1e14), received + maximumFee)

Sabnock01: Will try to fix this.

Sabnock01: I implemented my own. ;)

Sabnock01: Actually `wdivup` will cause a rounding error and try to `transferFrom` one more wei than was approved. This is fixed with `wdiv`.

Sabnock01: In this example `2003807` is approved and `transferFrom`'d. `wdivup` will try to `transferFrom` `2003808`

image

packages/foundry/contracts/test/other/tether/TetherJoin.t.sol
```diff
@@ -0,0 +1,285 @@
...
+        if (!equal(network, LOCALHOST)) vm.createSelectFork(network);
+
+        if (vm.envOr(MOCK, true)) setUpMock();
+        else setUpHarness(network);
+
+        //... Users ...
+        user = address(0xdeadbeef);
+        other = address(2);
+        me = 0xb4c79daB8f259C7Aee6E5b2Aa729821864227e84;
+        tetherMultiSig = 0xC6CDE7C39eB2f0F0095F41570af89eFC2C1Ea828;
```
alcueca, line 78: Nice

packages/foundry/contracts/test/other/tether/TetherJoin.t.sol
```diff
@@ -0,0 +1,285 @@
...
+}
+
+contract WithFeesTest is WithFees {
+    function testJoinPullWithFees() public {
+        track("userBalance", tether.balanceOf(user));
+        track("storedBalance", join.storedBalance());
+        track("joinBalance", tether.balanceOf(address(join)));
+
+        uint256 units = unit * 5;
+        uint256 fee = tether.basisPointsRate() * 100;
```
alcueca, line 183: Is this right?

You are using `fee` as a decimal factor, to multiply amounts by. Below you have `units * unit / (unit - fee)`, so `fee` needs to have the same decimals as `unit`. That means that you need to convert from `basisPointsRate` which is an fp4 (fixed point with 4 decimals) to `unit` which is a fixed point with `tether.decimals()`.

Wouldn't it be:
`uint256 fee = tether.basisPointsRate() * 10000;`

packages/foundry/contracts/test/other/tether/TetherJoin.t.sol
```diff
@@ -0,0 +1,285 @@
...
+        uint256 feeAdjustedUnits = units * unit / (unit - fee); // scale up first so precision isn't lost
```
alcueca, line 184: Here you are subtracting the fee. Below the `tether.approve(address(join), feeAdjustedUnits);` should be for `units`, which is the amount that is expected to leave the user's address.

packages/foundry/contracts/test/other/tether/TetherJoin.t.sol
```diff
@@ -0,0 +1,285 @@
...
+
+        vm.prank(user);
+        tether.approve(address(join), feeAdjustedUnits);
+        vm.prank(ladle);
+        join.join(user, uint128(feeAdjustedUnits));
```
alcueca, line 189: I might be getting everything wrong, but I would test in a more straightforward manner:
```
uint amount =  unit * 5;
uint fee =  amount * tether.basisPointsRate() / 10000; // Tether implementation, without SafeMath

vm.prank(user);
tether.approve(address(join), amount);
vm.prank(ladle);
join.join(user, uint128(amount));

assertTrackMinusEq("userBalance", amount, tether.balanceOf(user));
assertTrackPlusEq("storedBalance", amount - fee, join.storedBalance());
assertTrackPlusEq("joinBalance", amount - fee, tether.balanceOf(address(join)));
```

packages/foundry/contracts/test/other/tether/TetherJoin.t.sol
```diff
@@ -0,0 +1,285 @@
...
+        uint256 fee = tether.basisPointsRate() * 100;
+        uint256 feeAdjustedUnits = units * unit / (unit - fee); // scale up first so precision isn't lost
+
+        vm.prank(user);
+        tether.transfer(address(join), feeAdjustedUnits);
+        vm.prank(ladle);
+        join.join(user, uint128(units));
+
+        assertTrackMinusEq("userBalance", feeAdjustedUnits, tether.balanceOf(user));
+        assertTrackPlusEq("storedBalance", units, join.storedBalance());
```
alcueca, line 211: Here the fees would have been subtracted prior to calling `join.join()`, so you can skip the fee calculation and instead track the `tether.balanceOf(address(join))` as `amountReceived`. That tells you how much the `join` received (because in the harness it might already have some unaccounted balance).

Then your only assert is:
```
assertTrackPlusEq("storedBalance", amountReceived, join.storedBalance());
```

packages/foundry/contracts/test/other/tether/TetherJoin.t.sol
```diff
@@ -0,0 +1,285 @@
...
+        uint256 feeAdjustedUnits = units / 2 * unit / (unit - fee); // scale up first so precision isn't lost
+
+        vm.prank(user);
+        tether.approve(address(join), feeAdjustedUnits);
+        vm.prank(user);
+        tether.transfer(address(join), feeAdjustedUnits);
+        vm.prank(ladle);
+        join.join(user, uint128(units));
+
+        assertTrackMinusEq("userBalance", feeAdjustedUnits * 2, tether.balanceOf(user));
```
alcueca, line 231: Here your asserts are more complicated. Let's make a test where the user spends `amount * 2`.

 - Have the user to send an `amount` and obtain `amountReceived` by checking the change on `tether.balanceOf(address(join))`. You also obtain `feeCharged` as the difference.
 - Ask for  `join.join(amountReceived * 2)`. Since the `join` pulls to receive exactly what it is told, and it already has `amountReceived`, it will aim to obtain another `amountReceived` and the reverse fee formula should mean the user transfers `amount`.

Now you can check:
 - The user tether balance dropped by `amount * 2`.
 - The join tether balance increased by `amountReceived * 2`
 - The join stored balance increased by `amountReceived * 2`

This is also a very nice test, because we get to test our reverse fee implementation against tether's fee implementation, with no assumptions on the tests as to what the formulas actually are.

alcueca (changes requested):
It seems strange to me that you don't need `wdivup`. I'm not sure we are testing it right, so let's see if that needs fixing, and maybe then I'll be convinced.

This has turned into a very nice exercise on math and rounding. I think it is time well spent.

alcueca:
Thinking about it, I might add this job of reversing the tether fee formula into our solidity mentorship program. It's great.

alcueca:
Added the two simpler tests, and left the combined one for you. You might be right about the rounding after all 🤔

alcueca (approved):

PR #520, opened 2023-02-03
Update `ILadle` functions to be `payable`

Sabnock01:
`ILadle.sol` functions were missing `payable` modifier which made it unable to specify a `value` when calling them.

Also the module tests used a now deprecated `ILadleCustom` interface from when the module functions were missing from a previous version of `ILadle`. This has been fixed. 

As an aside, I've been using `forge fmt` on my commits lately which we should probably integrate into the CI. (That's why there's so many spacing changes)


alcueca (approved):
Thanks for that, I didn't know that the `payable` modifier belongs in the interfaces.

PR #522, opened 2023-02-06
Fix most compiler warnings

Sabnock01:
Mostly adds the correct function modifier, gets rid of unused vars, shadowed parameters.


alcueca (approved):
Thanks for fixing this

PR #526, opened 2023-02-20
`VRWitch`

iamsahu:


ultrasecreth (changes requested):
I truly believe that using inheritance like this is a no-no, better to extract the common functions into a lib and have sister witches

alcueca:
I have some blame here. The initial implementation was a completely unrelated Witch, that looked very similar. I suggested to @iamsahu to use inheritance to make peer reviews easier, not realizing that using different datatypes would mean the inheritance would be very messy.

I'm not even sure if having libraries with common functions would work that well. Few functions are actually 100% common. Maybe it makes sense to roll back to the previous implementation and having two completely unrelated witches with similar code.

alcueca:
For context, @ultrasecreth, [this was the original implementation without inheritance](https://github.com/yieldprotocol/vault-v2/pull/526/commits/11ac34d9271a2fb3bcef275abd53391bde3142a8#diff-869535c026252e8a05a06295fdbbbb0216dcc12ab4e95d9a1ec279dbd44ff789).

ultrasecreth (changes requested):
I believe this is not the right way, majority if not all the code duplication can be eliminated by using the "template method" pattern.
Also having a lib & a base class adds more complexity for no reason, the base class makes sense cause it deals with a lot of common state, so the lib is redundant imo, the lib made sense in a context of 2 sibling contracts with no common ancestor

ultrasecreth (commented):
I may be missing something, but I'm pretty sure that `WitchBase` can have like 90% of the code and then each concrete `Witch` would be just a few lines of code, but instead I see a lot of duplication that seems really unnecessary to me

packages/foundry/contracts/variable/VRWitch.sol
```diff
@@ -0,0 +1,152 @@
...
+            VRDataTypes.Vault memory vault
+        )
+    {
+        // If the world has not turned to ashes and darkness, auctions will malfunction on
+        // the 7th of February 2106, at 06:28:16 GMT
+        // TODO: Replace this contract before then 😰
+        // UPDATE: Enshrined issue in a folk song that will be remembered ✅
+        if (block.timestamp > type(uint32).max) {
+            revert WitchIsDead();
+        }
```
ultrasecreth, line 43: this should go to a common method or maybe even better, a modifier

packages/foundry/contracts/variable/VRWitch.sol
```diff
@@ -0,0 +1,152 @@
...
+    /// @dev Moves the vault ownership to the witch.
+    /// Useful as a method so it can be overridden by specialised witches that may need to do extra accounting or notify 3rd parties
+    /// @param vaultId Id of the vault to liquidate
+    function _auctionStarted(
+        bytes12 vaultId,
+        DataTypes.Auction memory auction_,
+        DataTypes.Line memory line
+    ) internal virtual returns (VRDataTypes.Vault memory vault) {
+        // The Witch is now in control of the vault under auction
+        vault = IVRCauldron(address(cauldron)).give(vaultId, address(this));
```
ultrasecreth, line 68: if you move this line out and make the function void, you can avoid duplicating this function

packages/foundry/contracts/variable/VRWitch.sol
```diff
@@ -0,0 +1,152 @@
...
+        )
+    {
+        DataTypes.Auction memory auction_ = auctions[vaultId];
+        VRDataTypes.Vault memory vault = IVRCauldron(address(cauldron)).vaults(
+            vaultId
+        );
+
+        // If the vault hasn't been auctioned yet, we calculate what values it'd have if it was started right now
+        if (auction_.start == 0) {
+            DataTypes.Balances memory balances = cauldron.balances(vaultId);
```
ultrasecreth, line 131: if you had a template method that returns the values you get from the vault and/or series, then you can avoid duplicating this whole method. You may also be able to re-use that method to simplify more code and/or remove even more duplication on other places

packages/foundry/contracts/WitchBase.sol
```diff
@@ -0,0 +1,680 @@
...
+    )
+        external
+        view
+        virtual
+        returns (
+            uint256 liquidatorCut,
+            uint256 auctioneerCut,
+            uint256 artIn
+        )
+    {}
```
ultrasecreth, line 624: I don't believe you need this, each witch will be forced to have it as it should be on each one of the interfaces

packages/foundry/contracts/WitchBase.sol
```diff
@@ -0,0 +1,680 @@
...
+        virtual
+        returns (uint256 artIn)
+    {}
+
+    /// @notice Returns base that could be paid given the artIn
+    function _debtToBase(DataTypes.Auction memory auction_, uint128 artIn)
+        internal
+        virtual
+        returns (uint256 baseIn)
+    {}
```
ultrasecreth, line 439: exactly what I was talking about :)

packages/foundry/contracts/WitchBase.sol
```diff
@@ -0,0 +1,680 @@
...
+    event LineSet(
+        bytes6 indexed ilkId,
+        bytes6 indexed baseId,
+        uint32 duration,
+        uint64 vaultProportion,
+        uint64 collateralProportion
+    );
+    event LimitSet(bytes6 indexed ilkId, bytes6 indexed baseId, uint128 max);
+    event ProtectedSet(address indexed value, bool protected);
+    event AuctioneerRewardSet(uint256 auctioneerReward);
```
ultrasecreth, line 69: Something I recently learned is that it's better to have errors and events in the interface, 2 main reasons
- the ABI of the interface is self-sufficient
- you can split the interface in 2 (one for events and other for methods & errors) and make the tests extend the former, so you avoid duplicating events on the tests and al de headaches that that creates

ultrasecreth (changes requested):

packages/foundry/contracts/Witch.sol
```diff
@@ -192,64 +28,28 @@ contract Witch is AccessControl {
     /// @return vault Vault that's being auctioned
     /// @return series Series for the vault that's being auctioned
     function auction(bytes12 vaultId, address to)
-        external
+        public
+        virtual
+        beforeAshes
```
ultrasecreth, line 33: 😆

ultrasecreth (approved):

PR #528, opened 2023-02-20
Remove hardhat, use only foundry imports with no remappings

alcueca:
Exactly that. It ruins the github tab on VSCode that now opens 22 connections, but it simplifies things a lot.


.gitmodules
```diff
@@ -1,15 +1,21 @@
...
+	path = lib/dss-interfaces
 	url = https://github.com/makerdao/dss-interfaces
-[submodule "packages/foundry/lib/forge-std"]
-	path = packages/foundry/lib/forge-std
+[submodule "lib/forge-std"]
+	path = lib/forge-std
 	url = https://github.com/foundry-rs/forge-std
-[submodule "packages/foundry/lib/yieldspace-tv"]
-	path = packages/foundry/lib/yieldspace-tv
+[submodule "/Users/alcueca/Code/vault-v2/lib/forge-std"]
```
ultrasecreth, line 10: pretty sure you don't want an absolute path here, prob a foundry bug if they added it this way

contracts/Cauldron.sol
```diff
@@ -3,16 +3,10 @@ pragma solidity >=0.8.13;
...
-import "@yield-protocol/utils-v2/contracts/math/WDivUp.sol";
-import "@yield-protocol/utils-v2/contracts/cast/CastU128I128.sol";
-import "@yield-protocol/utils-v2/contracts/cast/CastI128U128.sol";
-import "@yield-protocol/utils-v2/contracts/cast/CastU256U128.sol";
-import "@yield-protocol/utils-v2/contracts/cast/CastU256U32.sol";
-import "@yield-protocol/utils-v2/contracts/cast/CastU256I256.sol";
 import "./constants/Constants.sol";
+import "../lib/yield-utils-v2/src/access/AccessControl.sol";
+import "../lib/yield-utils-v2/src/utils/Math.sol";
+import "../lib/yield-utils-v2/src/utils/Cast.sol";
```
ultrasecreth, line 9: why no remappings?

ultrasecreth (changes requested):

contracts/Cauldron.sol
alcueca: Because I want to use submodules in the `environments-v2` repo

alcueca:
Will try to remove hardhat at a later point again

PR #529, opened 2023-02-23
Updated yield-utils

alcueca:
Exported via npm as 0.18.10.

Might remove hardhat for 0.19.0


Sabnock01 (approved):
Everything seems to be in order. I suppose since the Ladle and Cauldron are non-upgradable anyway it doesn't matter if we don't have the unit tests in Foundry yet, but my dev OCD doesn't like it. 

In any case after this we should be able to do away with `packages/` entirely, no?

alcueca:
Yes, doing away with `packages` is next. I might just go and do it.

And yes, not having unit tests is a bit icky. But my intention is to code them next time we have to deploy a Cauldron or Ladle. Right now we are maintaining an ugly codebase for no real gain.

PR #531, opened 2023-02-24
Contango wand

alcueca:
Still a draft. I've used copilot quite heavily, and there is a certain risk to this contract. We will need at least an internal audit.


packages/foundry/contracts/utils/ContangoWand.sol
```diff
@@ -0,0 +1,208 @@
...
+        compositeOracle = compositeOracle_;
+    }
+
+    /// ----------------- Cauldron Governance -----------------
+
+    /// @notice Copy the spotOracle and ratio from the master cauldron
+    function copySpotOracle(bytes6 baseId, bytes6 ilkId) external auth {
+        DataTypes.SpotOracle memory spotOracle_ = masterCauldron.spotOracles(baseId, ilkId);
+        cauldron.setSpotOracle(baseId, ilkId, spotOracle_.oracle, spotOracle_.ratio);
+    }
```
ultrasecreth, line 51: we'll prob never use this, but good to have

packages/foundry/contracts/utils/ContangoWand.sol
```diff
@@ -0,0 +1,208 @@
...
+    function copyLendingOracle(bytes6 baseId) external auth {
+        IOracle lendingOracle_ = masterCauldron.lendingOracles(baseId);
+        cauldron.setLendingOracle(baseId, lendingOracle_);
+    }
+
+    /// @notice Copy the debt limits from the master cauldron
+    function copyDebtLimits(bytes6 baseId, bytes6 ilkId) external auth {
+        DataTypes.Debt memory debt_ = masterCauldron.debt(baseId, ilkId);
+        cauldron.setDebtLimits(baseId, ilkId, debt_.max, debt_.min, debt_.dec);
+    }
```
ultrasecreth, line 63: ditto

packages/foundry/contracts/utils/ContangoWand.sol
```diff
@@ -0,0 +1,208 @@
...
+
+    /// @notice Add a new asset in the Cauldron, as long as it is an asset or fyToken known to the Yield Cauldron
+    function addAsset(bytes6 assetId) external auth {
+        address asset_ = masterCauldron.assets(assetId);
+        require(
+            asset_ != address(0) ||
+            address(masterCauldron.series(assetId).fyToken) != address(0),
+        "Asset not known to the Yield Cauldron");
+        cauldron.addAsset(assetId, asset_);
```
ultrasecreth, line 72: This won't work for a fyToken, it'd add `address(0)` as they're not assets on the master cauldron

packages/foundry/contracts/utils/ContangoWand.sol
```diff
@@ -0,0 +1,208 @@
...
+import "../interfaces/ILadle.sol";
+import "../interfaces/IJoin.sol";
+import "../oracles/yieldspace/YieldSpaceMultiOracle.sol";
+import "../oracles/composite/CompositeMultiOracle.sol";
+import "@yield-protocol/yieldspace-tv/src/interfaces/IPool.sol";
+import "@yield-protocol/utils-v2/src/access/AccessControl.sol";
+
+/// @title A contract that allows configuring the cauldron and ladle within bounds
+contract ContangoWand is AccessControl {
+    ICauldronGov public immutable cauldron;
```
ultrasecreth, line 16: we should call this `contangoCauldron` so the rest of the code is easier to understand

packages/foundry/contracts/utils/ContangoWand.sol
```diff
@@ -0,0 +1,208 @@
...
+    ICauldron public immutable masterCauldron;
+    ILadleGov public immutable ladle;
```
ultrasecreth, line 18: ditto

packages/foundry/contracts/utils/ContangoWand.sol
```diff
@@ -0,0 +1,208 @@
...
+        }
+        require(ratio_ >= bound_, "Ratio out of bounds");
+        cauldron.setSpotOracle(baseId, ilkId, compositeOracle, ratio_);
+    }
+
+    function _getDebtDecimals(bytes6 baseId, bytes6 ilkId) internal view returns (uint8 dec) {
+            // If the debt is already set in the cauldron, we use the decimals from there
+            // Otherwise, we use the decimals of the base
+            DataTypes.Debt memory cauldronDebt_ = ICauldron(address(cauldron)).debt(baseId, ilkId);
+            if (cauldronDebt_.sum != 0) {
```
ultrasecreth, line 112: you should prob check `cauldronDebt_.dec != 0` instead

packages/foundry/contracts/utils/ContangoWand.sol
```diff
@@ -0,0 +1,208 @@
...
+            max: max,
+            min: min,
+            dec: 0,
+            sum: 0
+        });
+    }
+
+    /// @notice Set the debt limits for a given asset pair in the Cauldron, within bounds
+    function setDebtLimits(bytes6 baseId, bytes6 ilkId, uint96 max, uint24 min) external auth {
+        // If the ilkId is a series and boundaries are not set, set them to default values
```
ultrasecreth, line 141: this comment should prob go before the `if`

packages/foundry/contracts/utils/ContangoWand.sol
```diff
@@ -0,0 +1,208 @@
...
+    }
+
+    /// @notice Propagate a token to the Ladle from the Yield Ladle
+    function addToken(address token) external auth {
+        ladle.addToken(token, masterLadle.tokens(token));
+    }
+
+    /// @notice Add join to the Ladle.
+    /// @dev These will often be used to hold fyToken, so it doesn't seem possible to put boundaries. However, it seems low risk. Famous last words.
+    function addJoin(bytes6 assetId, address join) external auth {
```
ultrasecreth, line 205: so the `assetId` of a `fyToken` always match the `seriesId`, so you could load the series from the `masterCauldron` and check that `address(series.fyToken) == IJoin(join).base()`.
I'd actually make the second param of type `IJoin` to avoid the casting.

in any case, this way you can be sure that the added `Join` is at least meant to hold the right token

ultrasecreth (commented):
The witch governance is missing, but's a really good start!

packages/foundry/contracts/utils/ContangoWand.sol
ultrasecreth: Also I'm thinking we could also have a factory method that deploys the `Join` and give the appropriate permissions, as Joins for fyTokens are never created as part of Yield's roll

alcueca:
image
We just need to add a `require(yieldLadle.joins(assetId) == IJoin(address(0)), "Join already exists")`

src/other/contango/ContangoWand.sol
```diff
@@ -0,0 +1,317 @@
...
+
+        AccessControl(address(join)).grantRole(IJoin.join.selector, address(contangoWitch));
+        AccessControl(address(join)).grantRole(IJoin.exit.selector, address(contangoWitch));
+    }
+
+    // TODO Maybe check if the join exists before deploying?
+
+    function deployJoin(bytes6 assetId) external auth returns (IJoin join) {
+        address asset = contangoCauldron.assets(assetId);
+        require(asset != address(0), "Asset not known to the Contango Cauldron");
```
alcueca, line 260: `require(yieldLadle.joins(assetId) == IJoin(address(0)), "Join already exists")`.

src/other/contango/ContangoWand.sol
```diff
@@ -0,0 +1,317 @@
...
+        _addJoin(assetId, join);
+    }
+
+    /// ----------------- Witch Governance -----------------
+
+    function setWitchDefaults(uint32 duration, uint64 vaultProportion, uint64 intialDiscount) external auth {
+        witchDefaults = WitchDefaults(duration, vaultProportion, intialDiscount);
+    }
+
+    function setLineAndLimit(
```
alcueca, line 276: What's the point of this function?

src/other/contango/ContangoWand.sol
```diff
@@ -0,0 +1,317 @@
...
+            baseId: baseId,
+            duration: witchDefaults.duration,
+            vaultProportion: witchDefaults.vaultProportion,
+            collateralProportion: uint64((WAD + witchDefaults.intialDiscount).wdivup(uint256(spotOracle_.ratio) * 10 ** 12)),
+            max: max
+        });
+    }
+
+    function setAuctioneerReward(uint256 auctioneerReward) external auth {
+        contangoWitch.setAuctioneerReward(auctioneerReward);
```
alcueca, line 301: There should be a `boundAuctioneerReward` function, I guess.

src/other/contango/ContangoWand.sol
```diff
@@ -0,0 +1,317 @@
...
+        bytes6 baseId,
+        uint32 duration,
+        uint64 vaultProportion,
+        uint64 collateralProportion,
+        uint128 max
+    ) external auth {
+        contangoWitch.setLineAndLimit(ilkId, baseId, duration, vaultProportion, collateralProportion, max);
+    }
+
+    function configureWitch(bytes6 ilkId, bytes6 baseId, uint128 max) external auth {
```
alcueca, line 287: I guess this one should be named `setLineAndLimit`

src/other/contango/ContangoWand.sol
```diff
@@ -0,0 +1,317 @@
...
+        contangoLadle.addPool(seriesId, pool_);
+    }
+
+    /// @notice Propagate an integration to the Ladle from the Yield Ladle
+    function addIntegration(address integration) external auth {
+        contangoLadle.addIntegration(integration, yieldLadle.integrations(integration));
+    }
+
+    /// @notice Propagate a token to the Ladle from the Yield Ladle
+    function addToken(address token) external auth {
```
alcueca, line 232: I like `copyX` for when we are propagating configs from the yield ladle to the contango ladle, as you did below.

src/other/contango/ContangoWand.sol
```diff
@@ -0,0 +1,317 @@
...
+    /// @notice Set the YieldSpace oracle as the source for a given asset pair in the Composite oracle, provided the source is set in the YieldSpace oracle
+    function setCompositeOracleSource(bytes6 baseId, bytes6 ilkId) external auth {
+        DataTypes.Series memory series_ = yieldCauldron.series(ilkId);
+        if (series_.fyToken != IFYToken(address(0))) {
+            (IPool pool_,) = yieldSpaceOracle.sources(baseId, ilkId);
+            require(address(pool_) != address(0), "YieldSpace oracle not set");
+            compositeOracle.setSource(baseId, ilkId, yieldSpaceOracle);
+        } else if (yieldCauldron.assets(ilkId) != address(0)) {
+            DataTypes.SpotOracle memory spotOracle_ = yieldCauldron.lookupSpotOracle(baseId, ilkId);
+            compositeOracle.setSource(baseId, ilkId, spotOracle_.oracle);
```
alcueca, line 203: Only if it is not `IOracle(address(0))`, I think.

src/other/contango/ContangoWand.sol
```diff
@@ -0,0 +1,317 @@
...
+        asset_ = yieldCauldron.assets(assetId);
+        if (asset_ == address(0)) {
+            asset_ = address(yieldCauldron.series(assetId).fyToken);
+        }
+        require(asset_ != address(0), "Asset not known to the Yield Cauldron");
+        contangoCauldron.addAsset(assetId, asset_);
+    }
+
+    /// @notice Add a new series, if it exists in the Yield Cauldron
+    function addSeries(bytes6 seriesId) external auth returns (DataTypes.Series memory series_) {
```
alcueca, line 103: As below, I think this should be `copySeries`

src/test/other/contango/ContangoWand.t.sol
```diff
@@ -0,0 +1,616 @@
...
+
+        bytes4 root = 0x0;
+        wand.grantRole(root, addresses[ARBITRUM][TIMELOCK]);
+
+        vm.startPrank(addresses[ARBITRUM][TIMELOCK]);
+
+        AccessControl(address(yieldLadle.joins(USDT))).grantRole(root, address(wand));
+        AccessControl(address(yieldCauldron.series(FYUSDT2306).fyToken)).grantRole(root, address(wand));
+
+        wand.grantRole(wand.copySpotOracle.selector, address(this));
```
alcueca, line 54: You know that there is a `grantRoles` function, right?

alcueca (commented):
Only reviewed the code, not the tests yet.

src/other/contango/ContangoWand.sol
ultrasecreth: good point

src/other/contango/ContangoWand.sol
ultrasecreth: have more granular access to the witch

src/other/contango/ContangoWand.sol
ultrasecreth: to set a min or max?

src/other/contango/ContangoWand.sol
ultrasecreth: it was originally overloaded with that name, but that doesn't play well when you wanna get the selector to assign permissions

src/other/contango/ContangoWand.sol
ultrasecreth: yeah, most of the names were there and I didn't understand the pattern, so didn't change them

src/other/contango/ContangoWand.sol
ultrasecreth: the lookup method does that already

src/other/contango/ContangoWand.sol
ultrasecreth: ditto

src/test/other/contango/ContangoWand.t.sol
ultrasecreth: yes, but creating the array is not as pretty as I'd like, and this way I can group the related permissions (wand + cauldron or whatever)

ultrasecreth:
BTW, the PR seems to fail cause it requires to fork arbitrum, I have the feeling the secret with the URL is missing in the repo config (or misnamed)

src/other/contango/ContangoWand.sol
ultrasecreth: btw, here I'm trying to avoid overriding an existing join, but I agree we should also check there isn't one that we could copy from the yieldLadle

PR #537, opened 2023-03-07
Identity Oracle

iamsahu:
An identity oracle which returns an exchange rate of 1:1 taking decimals into consideration for the conversion

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@yield-protocol/vault-v2",
-  "version": "0.18.12",
+  "version": "0.18.13",
   "description": "Yield Collateralized Debt Engine v2",
   "author": "Yield Inc.",
   "files": [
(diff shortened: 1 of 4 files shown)


ultrasecreth:
If it's making any modification, then it's not identity anymore 😉

alcueca (approved):
Thanks!

alcueca:
@ultrasecreth, calling it `IdentityMeaningPriceOfOneButAccountingForDecimals.sol` sounds a bit too wordy, I'm happy with the current name :P

ultrasecreth:
> @ultrasecreth, calling it `IdentityMeaningPriceOfOneButAccountingForDecimalsMultiOracle.sol` sounds a bit too wordy, I'm happy with the current name :P

Identity refers to the math property of a function that returns exactly the same as it's input, which this code no longer do.

This is more like a "pegged" oracle or whatever, but it's def not an identity anymore 🤓

PR #547, opened 2023-05-03
Variable rate AuditFixes

iamsahu:


cache/solidity-files-cache.json
```diff
@@ -1283,7 +1283,7 @@
       }
     },
     "lib/yieldspace-tv/lib/yield-utils-v2/src/token/ERC20.sol": {
-      "lastModificationDate": 1680017327438,
+      "lastModificationDate": 1681829815435,
```
Sabnock01, line 1286: We normally commit this file?

remappings.txt
```diff
@@ -3,4 +3,4 @@
 forge-std/=lib/forge-std/
 erc3156/=lib/ERC3156/
 dss-interfaces/=lib/dss-interfaces/
-openzeppelin/=lib/openzeppelin-contracts/
+@openzeppelin/=lib/openzeppelin-contracts/
```
Sabnock01, line 6: Why do we need `node` convention here?

src/variable/VRLadle.sol
```diff
@@ -338,20 +318,19 @@ contract VRLadle is UUPSUpgradeable, AccessControl() {
...
-            if (ink < 0) ilkJoin.exit(to, uint128(-ink));
+            else ilkJoin.exit(to, uint128(-ink));
         }
 
         // Manage base
         if (base != 0) {
             IJoin baseJoin = getJoin(vault.baseId);
             if (base < 0) baseJoin.join(vault.owner, uint128(-base));
-            if (base > 0) baseJoin.exit(to, uint128(base));
+            else baseJoin.exit(to, uint128(base));
```
Sabnock01, line 333: What if `base` is 0? That wasn't being checked before but is now?

src/variable/VRLadle.sol
```diff
@@ -297,26 +297,6 @@ contract VRLadle is UUPSUpgradeable, AccessControl() {
 
     // ---- Asset and debt management ----
 
-    /// @dev Move collateral and debt between vaults.
-    function stir(
```
Sabnock01, line 301: Why was `stir` moved exactly?

src/test/variable/VYToken.t.sol
```diff
@@ -259,75 +255,3 @@ contract VYTokenTest is VYTokenZeroState {
         assertLe(principalAmount, vyToken.convertToPrincipal(INK));
     }
 }
-
-contract FlashLoanEnabledStateTests is FlashLoanEnabledState {
```
Sabnock01, line 263: Why remove all these tests exactly?

src/test/variable/VRLadle.t.sol
```diff
@@ -154,18 +154,6 @@ contract VaultTests is VaultBuiltState {
         ladle.give(vaultId, admin);
     }
 
-    function testOnlyOwnerCouldMove() public {
```
Sabnock01, line 157: Why remove so many tests from this file? Is it because they're already tested elsewhere?

Sabnock01 (changes requested):

src/test/variable/VYToken.t.sol
iamsahu: Since, flash loan was removed the tests are removed as well.

src/test/variable/VRLadle.t.sol
iamsahu: The tests related to `stir` were removed.

src/variable/VRLadle.sol
iamsahu: On 325 we are checking that `base != 0`

cache/solidity-files-cache.json
iamsahu: Its getting commited since: https://github.com/yieldprotocol/vault-v2/commit/22d8fdd4d12cd4c40b90adbfccfb82275a458259

src/WitchBase.sol
```diff
@@ -49,7 +49,8 @@ contract WitchBase is AccessControl, IWitchEvents, IWitchErrors {
     mapping(bytes6 => mapping(bytes6 => DataTypes.Line)) public lines;
     mapping(bytes6 => mapping(bytes6 => DataTypes.Limits)) public limits;
     mapping(address => bool) public protected;
-
+    uint256[10] __gap;
```
alcueca, line 54: Maybe add a line of natspec explaining what this is.

src/oracles/VariableInterestRateOracle.sol
```diff
@@ -60,16 +66,15 @@ contract VariableInterestRateOracle is IOracle, AccessControl, Constants {
         uint256 slope2,
         IJoin join
     );
+    event AccumulationUpdated(bytes6 indexed baseId, bytes6 indexed kind, uint256 accumulated, uint256 lastUpdateTimestamp, uint256 utilizationRate);
```
alcueca:
```suggestion
    event AccumulatorUpdated(bytes6 indexed baseId, bytes6 indexed kind, uint256 accumulated, uint256 lastUpdateTimestamp, uint256 utilizationRate);
```

src/oracles/VariableInterestRateOracle.sol
```diff
@@ -175,54 +182,58 @@ contract VariableInterestRateOracle is IOracle, AccessControl, Constants {
...
-                    rateParameters.slope1 +
-                    (utilizationRate - rateParameters.optimalUsageRate)
-                        .wmul(rateParameters.slope2)
-                        .wdiv(1e18 - rateParameters.optimalUsageRate);
+        // Calculate the total debt
+        uint128 totalDebt;
+        DataTypes.Debt memory debt_;
+
+        for (uint256 i = 0; i < rateParameters.ilks.length; i++) {
+            if (cauldron.ilks(base.b6(), rateParameters.ilks[i])) {
```
alcueca: You can remove this `if` and the code will work the same, but saving some gas.

src/variable/VRLadle.sol
```diff
@@ -217,7 +223,7 @@ contract VRLadle is UUPSUpgradeable, AccessControl() {
 
     /// @dev Accept Ether, wrap it and forward it to the provided address
     /// This function should be called first in a batch, and the Join should keep track of stored reserves
-    /// Passing the id for a join that doesn't link to a contract implemnting IWETH9 will fail
+    /// Passing the id for a join that doesn't link to a contract implementing IWETH9 will fail
```
alcueca: The natspec is outdated, we don't pass an asset id anymore.

src/variable/VRLadle.sol
alcueca: It doesn't update the global debt counters when `ilkId != baseId`. It could be fixed, but at the same time we haven't ever used it, so better to just reduce the attack surface.

alcueca (approved):
Just cosmetic changes required.

